PR #9, opened 2023-01-16
Add linters and GHA job to run them on PRs

pmacik:
Signed-off-by: Pavel Macík <pavel.macik@gmail.com>


sadlerap:
Would it be worth adding something like [shellcheck](https://github.com/koalaman/shellcheck) for the testing infrastructure we have?

pmacik:
> Would it be worth adding something like [shellcheck](https://github.com/koalaman/shellcheck) for the shell scripts we have?

Good suggestion, will take a look. Thx

sadlerap (commented):
Looks like yamllint is trying to lint our github actions infrastructure (`.github/workflows/pr-checks.yaml`); is it possible to disable checking for this particular file?  We're probably going to get more false positives than actual problems.

pmacik:
> Would it be worth adding something like [shellcheck](https://github.com/koalaman/shellcheck) for the shell scripts we have?

@sadlerap I've included the shellcheck, PTAL

pmacik:
> Looks like yamllint is trying to lint our github actions infrastructure (`.github/workflows/pr-checks.yaml`); is it possible to disable checking for this particular file? We're probably going to get more false positives than actual problems.

@sadlerap It can be disabled via [this](https://yamllint.readthedocs.io/en/stable/disable_with_comments.html#disabling-all-checks-for-a-file), however I'd like to have it checked as well as the other YAMLs. I don't see a reason why not to.

Anyway, I have tweaked yamllint configuration (to disable line-lenght check) as we usually ignore those kind of warnings. So with the current state it runs without errors and warnings.

dperaza4dustbit (approved):
/lgtm

.github/workflows/pr-checks.yaml
```diff
@@ -16,6 +16,49 @@ env:
...
+        if: success() || failure()
+        run: make lint-feature-files
+
+      - name: Lint conflicts in source files
+        if: success() || failure()
+        run: make lint-conflicts
+
+      - name: Lint shell scripts
+        if: success() || failure()
+        run: make lint-shell
```
sadlerap, line 60: Why keep these as separate steps?  `make lint` will do roughly the same things, so what's the difference?

sadlerap (approved):
LGTM, with one question.  I'm fine with it as is, just trying to understand it.

filariow (approved):

.github/workflows/pr-checks.yaml
pmacik: @sadlerap, you are correct, the `make lint` runs it all. However, it runs it all in one big chunk and if any of the linters fails, it is tedious to go through the large log to investigate. By splitting it into individual steps, it is immediately clear which one of the linters failed and much easier to investigate or debug just the relevant log.

pmacik: Additionally, by using the "split" approach all the linters are executed even if some of them failed, while in case of using `make lint`, the execution would fail fast as soon as any linter failed, actually skipping the remaining ones. So if multiple linters are set to fail, you would always see just the first one that faied, missing the others.

`make -k lint` would fix that but you would still have to go through a large overall log to find the actual problem.

PR #10, opened 2023-01-18
Changes to RegisteredService CRD to match design

dperaza4dustbit:
Adding HealthCheck section and modifying the SCI and SED sections. SED is also changed to allow for value to be a reference to a secret key.

Signed-off-by: David Peraza <dperaza@redhat.com>


api/v1alpha1/registeredservice_types.go
```diff
@@ -25,26 +25,67 @@ type RegisteredServiceConstraints struct {
     Environments []string `json:"environments,omitempty"`
 }
 
+type HealthCheckContainer struct {
+    // Container image with the client to run the test
+    Image string `json:"image"`
+    // Command to execute in the container to run the test
+    Cmd string `json:"cmd"`
```
sadlerap, line 32: nit: The struct member name _should_ match the documentation, like so:

```suggestion
    // Command to execute in the container to run the test
    Command string `json:"cmd"`
```

api/v1alpha1/registeredservice_types.go
```diff
@@ -25,26 +25,67 @@ type RegisteredServiceConstraints struct {
     Environments []string `json:"environments,omitempty"`
 }
 
+type HealthCheckContainer struct {
```
sadlerap, line 28: nit: Add documentation for this struct and the others you added.  It's useful in some scenarios (example: it's what's used for `kubectl explain`)

api/v1alpha1/registeredservice_types.go
```diff
@@ -25,26 +25,67 @@ type RegisteredServiceConstraints struct {
...
+    // Command to execute in the container to run the test
+    Cmd string `json:"cmd"`
+
+}
+
+type RegisteredServiceHealthCheck struct {
+    // Container defines a container that will run a check against the 
+    // ServiceEndpointDefinition to determine connectivity and access.
+    Container HealthCheckContainer `json:"container"`
+}
```
sadlerap, line 40: Long-term, it may be useful to make this a [`PodSpec` data structure](https://godocs.io/k8s.io/api/core/v1#PodSpec).  Example scenario where this may be useful: if the health check needs to consume some secrets to connect to the service to determine the service's status, we're going to need to add a way to support projecting secrets according to the users needs.

Rather than rebuild our own version of `PodSpec`, let's just expose the `PodSpec` data structure we're going to need to generate anyway for whichever pod runner we decide to use.

Thoughts?

sadlerap (commented):
As it stands, I'm fine with this PR landing as-is, though I have some nits & some thoughts on design going forward into the future.

api/v1alpha1/registeredservice_types.go
dperaza4dustbit: OK

api/v1alpha1/registeredservice_types.go
dperaza4dustbit: Got it

api/v1alpha1/registeredservice_types.go
dperaza4dustbit: Do we expose PodSpec directly in RegisteredServiceHealthCheck. I would say inside HealthCheckContainer better right To make it easy to add other healthcheck like TCP or http for example. Also by just adding container image and command we are saying just tell me the container and I will make sure internally I run it with a podspec I think will work for you. Since we are not even using healthcheck right now lets leave this as a placeholder and decide on this later when we implement health check

sadlerap: Yeah, let's just leave this for now; just wanted to start thinking about this.

sadlerap (dismissed):
LGTM

baijum (approved):

PR #20, opened 2023-02-01
docker build cleanups

sadlerap:
This fixes building with `make docker-buildx`, which was broken.  Docker was not sending the correct build context, since we were never informing it which directory it should use.

This also ignores `out/` directories during docker builds, which should speed up build times.  We don't need to send a python virtual environment & acceptance test outputs to docker when building; we solely depend on go code there.

Signed-off-by: Andy Sadler <ansadler@redhat.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,3 +2,6 @@
 # Ignore build and test binaries.
 bin/
 testbin/
+
+# ignore testing-related artifacts
+out/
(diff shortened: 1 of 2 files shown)


filariow (approved):

baijum (approved):

sadlerap:
Rebased.

filariow (approved):

PR #21, opened 2023-02-02
Add Application Agent

filariow:
Update the repo to also host the application agent's source code

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #23, opened 2023-02-02
Add Service Agent

filariow:
Needs #21 to be merged.

### DIFF
--- a/make/agents/svc/deploy.mk
+++ b/make/agents/svc/deploy.mk
@@ -1,5 +1,5 @@
 KUSTOMIZE_ARGS=--load-restrictor LoadRestrictionsNone
-KUSTOMIZE_OVERLAY=config/crd/overlays/agentapp
+KUSTOMIZE_OVERLAY=config/crd/overlays/agentsvc
 
 ifndef ignore-not-found
   ignore-not-found = false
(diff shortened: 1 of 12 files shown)


cmd/primaza/main.go
```diff
@@ -0,0 +1,136 @@
...
+		setupLog.Error(err, "unable to create controller", "controller", "ServiceClass")
+		os.Exit(1)
+	}
+	if err = (&controllers.RegisteredServiceReconciler{
+		Client: mgr.GetClient(),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "RegisteredService")
+		os.Exit(1)
+	}
```
sadlerap, line 119: These need to be moved to the service agent.

main.go
```diff
@@ -1,143 +0,0 @@
-/*
```
sadlerap, line 1: Why are we making `main.go` a symlink?  I don't see any advantage to doing so.

sadlerap (changes requested):
Please rebase; feedback also applies to #21.

Also, for everyone else: is there any reason why we can't merge only this, instead of having to merge both this and #21?  IMO it doesn't make sense to split this across multiple PRs, since splitting primaza into three separate controllers seems like something that should be happening across one PR, not two.  This seems like it's making a bit of extra work for not much benefit, even if we're splitting this across multiple commits.

main.go
filariow: to maintain compatibility with operator-sdk

cmd/primaza/main.go
filariow: you're right on ServiceClassReconciler. We should not need RegisteredServiceReconciler, because the agent will create RegisteredService on Primaza

sadlerap (dismissed):
LGTM

I'll base my service class stuff off of this.

make/agents/svc/deploy.mk
```diff
@@ -0,0 +1,14 @@
...
+  ignore-not-found = false
+endif
+
+.PHONY: install
+install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build $(KUSTOMIZE_ARGS) $(KUSTOMIZE_OVERLAY) | kubectl apply -f -
+
+.PHONY: uninstall
+uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build $(KUSTOMIZE_ARGS) $(KUSTOMIZE_OVERLAY) | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```
sadlerap, line 14: Q: how is a developer intended to deploy an agent for testing purposes?  I don't see a deploy rule defined anywhere.

filariow: nice catch, added!

sadlerap (approved):
LGTM!

PR #25, opened 2023-02-03
Add support for split acceptance testing

sadlerap:
Running all acceptance tests currently requires them to be run on a single machine.  This change introduces the ability to split test runs into multiple externally managed runs.

To accomplish this, this change introduces two new environment variables that control which tests are run:
- RUNS defines how many runs to distribute testing across.  This must be greater than or equal to 1.
- RUN_ID assigns a numeric identifier to a particular run.  This must be in the range [0, RUNS-1]. Using these two variables, we can distribute the acceptance testing scenarios across each defined test run.

As an example, let's say we have two machines for running tests, A and B.  We set RUNS=2 on both machines and also set RUN_ID=0 on A and RUN_ID=1 on B.  When acceptance tests are run, A and B will each execute half of the defined acceptance testing scenarios.

By default, we set RUNS=1 and RUN_ID=0, so no test splitting is done without explicitly asking for it.

Signed-off-by: Andy Sadler <ansadler@redhat.com>

### DIFF
--- a/test/acceptance/features/environment.py
+++ b/test/acceptance/features/environment.py
@@ -16,8 +16,37 @@ before_all(context), after_all(context)
 from behave import fixture, use_fixture
 from steps.command import Command
 from steps.kind import KindProvider
+import os
 
 cmd = Command()
+
+
+class Runner(object):
+    def __init__(self, num_runners: int = 1, runner_id: int = 0):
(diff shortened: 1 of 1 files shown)


filariow:
superceded by #49

PR #30, opened 2023-02-14
Handle Registered Service delete event

dperaza4dustbit:
Handling delete via event triggered by delete timestamp change. No need to add a finalizer yet. Also fixed the duplicate entries in service catalog.

Signed-off-by: David Peraza <dperaza@redhat.com>

### DIFF
--- a/controllers/registeredservice_controller.go
+++ b/controllers/registeredservice_controller.go
@@ -36,6 +36,15 @@ type RegisteredServiceReconciler struct {
 	Scheme *runtime.Scheme
 }
 
+func ServiceInCatalog(sc primazaiov1alpha1.ServiceCatalog, serviceName string) int {
+	for i, service := range sc.Spec.Services {
+		if service.Name == serviceName {
+			return i
+		}
+	}
+	return -1
+}
+
 //+kubebuilder:rbac:groups=primaza.io,resources=registeredservices,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=primaza.io,resources=registeredservices/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=primaza.io,resources=registeredservices/finalizers,verbs=update
@@ -56,8 +65,37 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	log := log.FromContext(ctx)
 
 	var rs primazaiov1alpha1.RegisteredService
-	if err := r.Client.Get(ctx, req.NamespacedName, &rs); err != nil {
-		log.Error(err, "Error fetching RegiteredService")
+	err := r.Client.Get(ctx, req.NamespacedName, &rs)
+	if err != nil && errors.IsNotFound(err) {
+		log.Info("Registered Service not found, handling delete event")
+		var sc primazaiov1alpha1.ServiceCatalog
+		err = r.Get(ctx, types.NamespacedName{
+			Name:      "primaza-service-catalog",
+			Namespace: req.NamespacedName.Namespace, // "primaza-system",
+		}, &sc)
+		if err != nil {
+			log.Error(err, "Error fetching Service Catalog")
+			return ctrl.Result{}, err
+		}
+
+		si := ServiceInCatalog(sc, req.Name)
+
+		if si == -1 {
+			log.Info("Registered Service is being deleted but no catalog entry found")
+			return ctrl.Result{}, nil
+		}
+
+		sc.Spec.Services = append(sc.Spec.Services[:si], sc.Spec.Services[si+1:]...)
+		log.Info("Updating Service Catalog")
+		if err := r.Update(ctx, &sc); err != nil {
+			// Service Catalog update failed
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, nil
+
+	} else if err != nil {
+		log.Error(err, "Error fetching RegisteredService")
 		return ctrl.Result{}, err
 	}
 
@@ -82,7 +120,7 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	rs.Status = status
 
 	log.Info("Updating status of RegisteredService")
-	err := r.Status().Update(context.Background(), &rs)
+	err = r.Status().Update(ctx, &rs)
 	if err != nil {
 		log.Error(err, "RegisteredService Status Failed")
 		return ctrl.Result{}, err
@@ -135,7 +173,7 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		} else if err != nil {
 			// Error that isn't due to the ServiceCatalog not found
 			return ctrl.Result{}, err
-		} else {
+		} else if ServiceInCatalog(sc, scs.Name) == -1 {
 			log.Info("Updating Service Catalog")
 			sc.Spec.Services = append(sc.Spec.Services, scs)
 			err = r.Update(ctx, &sc)
--- a/test/acceptance/features/steps/kubernetescli.py
+++ b/test/acceptance/features/steps/kubernetescli.py
@@ -153,6 +153,16 @@ spec:
         assert exit_code == 0, f"Non-zero exit code ({exit_code}) while deleting a YAML: {output}"
         return output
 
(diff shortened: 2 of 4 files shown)


controllers/registeredservice_controller.go
```diff
@@ -56,7 +65,31 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			Name:      "primaza-service-catalog",
+			Namespace: req.NamespacedName.Namespace, // "primaza-system",
+		}, &sc)
+		if err != nil {
+			log.Error(err, "Error fetching Service Catalog")
+			return ctrl.Result{}, err
+		}
+
+		si := ServiceInCatalog(sc, req.Name)
+		sc.Spec.Services = append(sc.Spec.Services[:si], sc.Spec.Services[si+1:]...)
```
filariow, line 82: if the catalog does not contain the service, the function `ServiceInCatalog` returns `-1` and `sc.Spec.Services[:-1]`  will panic.

```go
a := []string{"a", "b", "c"}
si := -1
a = append(a[:si], a[si+1:]...)
fmt.Printf("a: %v\n", a)

---

panic: runtime error: slice bounds out of range [:-1]
```

controllers/registeredservice_controller.go
```diff
@@ -56,7 +65,31 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	log := log.FromContext(ctx)
 
 	var rs primazaiov1alpha1.RegisteredService
-	if err := r.Client.Get(ctx, req.NamespacedName, &rs); err != nil {
+	err := r.Client.Get(ctx, req.NamespacedName, &rs)
+	if err != nil && errors.IsNotFound(err) {
+		log.Info("Registered Service not found, hanlding delete event")
```
filariow, line 70:
```suggestion
		log.Info("Registered Service not found, handling delete event")
```

controllers/registeredservice_controller.go
```diff
@@ -56,7 +65,31 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		if err != nil {
+			log.Error(err, "Error fetching Service Catalog")
+			return ctrl.Result{}, err
+		}
+
+		si := ServiceInCatalog(sc, req.Name)
+		sc.Spec.Services = append(sc.Spec.Services[:si], sc.Spec.Services[si+1:]...)
+		log.Info("Updating Service Catalog")
+		err = r.Update(ctx, &sc)
+		if err != nil {
```
filariow, line 85:
```suggestion
		if err := r.Update(ctx, &sc); err != nil {
```

to stick to [Effective go guidelines](https://go.dev/doc/effective_go#if), we can use this `if`s format when possible. Obviously, that's not mandatory

filariow (changes requested):

controllers/registeredservice_controller.go
dperaza4dustbit: k, let me add a check before the remove. But reason test did not catch is that there is always a catalog entry if a registered service is being deleted.

controllers/registeredservice_controller.go
dperaza4dustbit: thanks

controllers/registeredservice_controller.go
dperaza4dustbit: OK

filariow (dismissed):

controllers/registeredservice_controller.go
```diff
@@ -56,7 +65,36 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		log.Info("Updating Service Catalog")
+		if err := r.Update(ctx, &sc); err != nil {
+			// Service Catalog update failed
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, nil
+
+	} else if err != nil {
 		log.Error(err, "Error fetching RegiteredService")
```
sadlerap, line 98:
```suggestion
		log.Error(err, "Error fetching RegisteredService")
```

controllers/registeredservice_controller.go
```diff
@@ -56,7 +65,36 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	log := log.FromContext(ctx)
 
 	var rs primazaiov1alpha1.RegisteredService
-	if err := r.Client.Get(ctx, req.NamespacedName, &rs); err != nil {
+	err := r.Client.Get(ctx, req.NamespacedName, &rs)
+	if err != nil && errors.IsNotFound(err) {
+		log.Info("Registered Service not found, handling delete event")
```
sadlerap, line 70: I'm not convinced this is right; AIUI if you want deletion events to show up in your controller, you need to set a finalizer.  See [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) for more details.

test/acceptance/features/steps/kubernetescli.py
```diff
@@ -153,6 +153,16 @@ def delete(self, yaml, namespace=None):
...
         return output
 
+    def delete_by_name(self, res_type, res_name, namespace=None):
+        if namespace is not None:
+            ns_arg = f"-n {namespace}"
+        else:
+            ns_arg = ""
+        (output, exit_code) = self.cmd.run(
+            f"{ctx.cli} delete {res_type} {res_name} {ns_arg}")
+        assert exit_code == 0, f"Non-zero exit code ({exit_code}) while deleting a Custome Resource: {output}"
```
sadlerap, line 163: Nit:
```suggestion
        assert exit_code == 0, f"Non-zero exit code ({exit_code}) while deleting a Custom Resource: {output}"
```

controllers/registeredservice_controller.go
```diff
@@ -82,7 +120,7 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	rs.Status = status
 
 	log.Info("Updating status of RegisteredService")
-	err := r.Status().Update(context.Background(), &rs)
+	err = r.Status().Update(context.Background(), &rs)
```
sadlerap, line 123: Don't make a new context here; use the context you get passed from controller-runtime.
```suggestion
	err = r.Status().Update(ctx, &rs)
```

controllers/registeredservice_controller.go
```diff
@@ -36,6 +36,15 @@ type RegisteredServiceReconciler struct {
 	Scheme *runtime.Scheme
 }
 
+func ServiceInCatalog(sc primazaiov1alpha1.ServiceCatalog, serviceName string) int {
+	for i := 0; i < len(sc.Spec.Services); i++ {
+		if sc.Spec.Services[i].Name == serviceName {
+			return i
+		}
+	}
+	return -1
```
sadlerap, line 45: Can simplify this slightly:
```suggestion
	for i, service := range sc.Spec.Services {
		if service.Name == serviceName {
			return i
		}
	}
	return -1
```

sadlerap (changes requested):
Looks mostly good, a few nits.

controllers/registeredservice_controller.go
dperaza4dustbit: This seems to work since i don't need data from the object, obviously I cannot get it cause that is how I get in the error state, but if you notice the req.name has the right name for the request. I was thinking this is an easier way to implement initially. If for some reason I need data in the resource being delete I have not choice but to implement a finilizer

controllers/registeredservice_controller.go
dperaza4dustbit: OK

test/acceptance/features/steps/kubernetescli.py
dperaza4dustbit: OK

controllers/registeredservice_controller.go
dperaza4dustbit: OK

controllers/registeredservice_controller.go
dperaza4dustbit: I like it, thanks

filariow (approved):

filariow (approved):

filariow (dismissed):

filariow (approved):

PR #28, opened 2023-02-10
Add tests for agents deployment

filariow:
Add most of BDD scenarios for [agents deployment story](https://issues.redhat.com/browse/APPSVC-1288). Updates tests code.

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #31, opened 2023-02-15
Agents deployment in linked clusters

filariow:
> Depends on #28 

[Story 1288](https://issues.redhat.com/browse/APPSVC-1288)

On Cluster Environment creation agents are pushed to its Service and Application namespaces.
When a namespace is added to a Cluster Environment Service or Application namespaces list, agents are pushed into it.
When a namespace is deleted from a Cluster Environment Service or Application namespaces list, agents are removed from it.
When a Cluster Environment is deleted, previously pushed agents are removed.

### DIFF
--- a/config/agents/svc/rbac/role.yaml
+++ b/config/agents/svc/rbac/role.yaml
@@ -17,6 +17,25 @@ rules:
   - patch
   - delete
 - apiGroups:
+  - primaza.io
+  resources:
+  - servicebindings
+  - serviceclaims
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
+- apiGroups:
   - ""
   resources:
   - events
--- a/pkg/primaza/constants/doc.go
+++ b/pkg/primaza/constants/doc.go
@@ -1,0 +1,18 @@
+/*
+Copyright 2023 The Primaza Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package constants contains Primaza's constants
+package constants
--- a/pkg/primaza/controlplane/doc.go
+++ b/pkg/primaza/controlplane/doc.go
@@ -1,0 +1,19 @@
+/*
+Copyright 2023 The Primaza Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package controlplane contains code for setting up and running Primaza
+// control plane
+package controlplane
--- a/pkg/primaza/workercluster/agentapp.go
+++ b/pkg/primaza/workercluster/agentapp.go
@@ -1,0 +1,146 @@
+/*
+Copyright 2023 The Primaza Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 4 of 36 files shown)


config/agents/svc/rbac/role.yaml
```diff
@@ -16,6 +16,25 @@ rules:
...
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
```
dperaza4dustbit, line 37: hum... do you need list, watch and patch? are you missing delete?

pkg/primaza/constants/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package constants contains Primaza's constants
+package constants
```
dperaza4dustbit, line 18: No much going on here, why do we need multiple constants files?

pkg/primaza/controlplane/doc.go
```diff
@@ -0,0 +1,19 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package controlplane contains code for setting up and running Primaza
+// control plane
+package controlplane
```
dperaza4dustbit, line 19: No much going on.

pkg/primaza/workercluster/agentapp.go
```diff
@@ -0,0 +1,225 @@
...
+
+	return nil
+}
+
+func PushApplicationAgent(ctx context.Context, cli *kubernetes.Clientset, namespace string) error {
+	if _, err := createAgentAppDeployment(ctx, cli, namespace); err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+
+	if _, err := createAgentAppService(ctx, cli, namespace); err != nil && !errors.IsAlreadyExists(err) {
```
dperaza4dustbit, line 48: why do we need a service for the agents?

pkg/primaza/workercluster/agentapp.go
```diff
@@ -0,0 +1,225 @@
...
+        control-plane: primaza-controller-agentapp
+    spec:
+      securityContext:
+        runAsNonRoot: true
+      containers:
+      - command:
+        - /manager
+        args:
+        - --leader-elect
+        image: agentapp:latest
```
dperaza4dustbit, line 191: hardcoded? is this temporary? Lets talk about the name of these images in scrum or refinement.

pkg/primaza/workercluster/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package workercluster contains code for interacting with a worker cluster
+package workercluster
```
dperaza4dustbit, line 18: Not clear to me why we have all these doc.go files

dperaza4dustbit (commented):
Interesting approach. I have some questions inline

pkg/primaza/constants/doc.go
filariow: the `doc.go` is one of the suggested approaches for [documenting packages](https://go.dev/blog/godoc)

config/agents/svc/rbac/role.yaml
filariow: that's the role that will be assigned to the service agent. it just need read access. I will remove update/patch.

pkg/primaza/workercluster/agentapp.go
filariow: yes, that's a placeholder for now and will be replaced in next PRs

pkg/primaza/workercluster/agentapp.go
filariow: you're right it's not needed, I added this debugging the problems related to leases and forgot it here. nice catch!

config/agents/svc/rbac/role.yaml
dperaza4dustbit: I still see the update and watch there

dperaza4dustbit (approved):
Since this PR is so big can we merge to get Andy unblocked and then we can do a demo in cabal where we can ask question, my main concern with this PR is giving primaza role to much access to deployments in the target namespaces. We should be able to create a deployment for the agent and maybe delete it, but there is no need to read, list or update any other deployments that could be from the applications

filariow:
@dperaza4dustbit I investigated permissions further and we can scope Get/List/Update/Patch/Delete permission with resource name, so that will be limited to agent deployment. We can't do that for Create, because in kubernetes RBAC happens at URL level. I updated Primaza's permission on namespace resources: https://github.com/primaza/primaza/blob/a8b238f5a0f882f7e3ae6ab60295fcf72566160d/test/acceptance/features/steps/workercluster.py#L165

dperaza4dustbit (approved):
/lgtm

PR #32, opened 2023-02-15
Remove minikube download from CI

filariow:


.github/actions/setup-cli/setup_tools.sh
```diff
@@ -16,11 +16,11 @@ if [ "$KUBECTL" == true ] || [ "$START_MINIKUBE" == true ]; then
     mv -v kubectl "$GITHUB_WORKSPACE/bin/"
 fi
 
-if [ "$MINIKUBE" == true ] || [ "$START_MINIKUBE" == true ]; then
-    echo "Downloading minikube..."
-    curl -Lo minikube "https://storage.googleapis.com/minikube/releases/v${MINIKUBE_VERSION}/minikube-linux-amd64"
-    chmod +x minikube
-    mv -v minikube "$GITHUB_WORKSPACE/bin/"
-fi
+# if [ "$MINIKUBE" == true ] || [ "$START_MINIKUBE" == true ]; then
```
baijum, line 19: I think we can remove it instead of commenting out the code.

filariow: yes, commenting out the code was the best idea. I changed the input parameter, so it should now just not be triggered. What do you think?

baijum (dismissed):

.github/actions/setup-cli/setup_tools.sh
baijum: LGTM

baijum (approved):

PR #34, opened 2023-02-16
Fix kubernetes API server crash in BDD tests

filariow:
Disabling anonymous access causes kubernetes API server to crash and the cluster to be unstable (see https://github.com/kubernetes/kubernetes/issues/43784).

This PR enables the anonymous access again and momentarily disables the test relying on this feature to be off.
The disabled test will be modified and enabled again in [Story 1272](https://issues.redhat.com/browse/APPSVC-1272).


mmulholla (commented):
LGTM. Tested locally with the change and it fixed my issue.

PR #40, opened 2023-02-21
Adding initial documentation for Registered Service

dperaza4dustbit:
Add documentaion describing the Registered Service resource.

### DIFF
--- a/docs/entities/registeredservice.md
+++ b/docs/entities/registeredservice.md
@@ -1,0 +1,68 @@
+# RegisteredService
+
+A RegisteredService represents an instance of a running service.
+This resource can be created by both a human operator or a piece of software (agent, bot, script). This flexibility allows for the service provisioning and service discovery operations to be decoupled from representation of that service in a service catalog.
+The service catalog can be consumed by Kubernetes applications that are unaware of the lifecycle of the service.
+The creation of a registered service signals an update to the service catalog.
+The service catalog can then be advertised to the application developers.
+A RegisteredService can be claimed to be used by a specific application, and once claimed, the RegisteredService cannot be claimed by other application.
+If the application does not require the service any longer, it can remove the claim and the registered service is put back in the pool of available services.
+
+## Specification
+
+The definition of a RegisteredService can be obtained directly from our [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). The specification contains two required properties:
+
+- ServiceClassIdentity: A set of key/value pairs that identify the service class. Examples of service class identity keys include type of service, and provider of service.
+This property is required.
+- ServiceEndpointDefinition: A set of key/value pairs that provides two pieces of information: the service connectivity and the service authentication.
+Values can be either a string or a reference to a secret field.
+This property is required.
+
+A RegisteredService also has three optional properties, which gives the user more control over the resource:
+
+- Constraints: Restrictions of the registered service.
+A constraint could be a specific environment where the service cannot be claimed from.
+As an example, a constraint could describe that I don't want this service to be used on any environment except for production.
+This property is optional, when it is absent, it means there is not constraints.
+- HealthCheck: A mechanism to be able to verify the service is online and ready to use.
+One way this can be accomplished is by providing an image containing a client that can be run to test connectivity and authentication. This property is optional, when it is absent, it means the service will be considered available as soon as it is registered.
+- SLA: Provides multiple levels of resiliency, scalability, fault tolerance and security. This allows claims to take into account the robustness of service. This property is optional, when it is absent, it means that there is no distinctions between services given the SLA.
+
+
+## Status
+
+The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml).
+It contains only one required property to track state.
+The state could be one of the following: registered, available, claimed, unreachable or undetermined.
+The registered service state is "registered" right after creation.
+In most cases this will change immediately after the controller gets notified of existence of new resource, unless controller fails even before it tries to run health check.
+If the health check passes or is not defined, the state will change to "available".
+Once a RegisteredService is claimed, its state moves to "claimed".
+On the other hand, if the health check comes back as a failure, the state would change to "unreachable".
+In some cases, the health check could fail before it can determine whether the service is healthy or not.
+In those cases, the state of the registered service will go to "undetermined" indicating there is an issue with the health check and not with the service itself.
+
+There are some situations when the health check will fail while the registered service is in "claimed" state.
+In those cases the registered service will move to "unreachable".
+If, at a later time, the health check passes then the controller will check if there is still a claim matching the registered service and move the state back to "claimed".
+However, if there is not claim matching the registered service the state will move to "available"
+
+## Use Cases
+
+### Creation
+
+When a RegisteredService is created, Primaza will first check if the health check can be run and will modify the state accordingly.
+The ServiceCatalog will be updated to include the new RegisteredService only if the state can be moved to "available".
+
+### Deletion
+
+When a RegisteredService resource is deleted, the ServiceCatalog entry for the service should be deleted.
+Also, if a RegisteredService is claimed, the ServiceClaim resource state will be changed to "Pending" by the ServiceClaim controller since the matched RegisteredService does not exist any longer.
+Additionally, when a RegisteredService resource state changes to "claimed" the corresponding entry in the ServiceCatalog resource is removed.
+
+### Update
+
+When a RegisteredService is updated, a few things can happen:
+- Existing ServiceClaim resources may not be able to match the registered service any longer.
+In those case the service claim controller will try to find another match and, if unsuccessful, the ServiceClaim resource state will change to "pending".
+- Existing service endpoint definition has changed and should trigger a binding secret generation. 
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Primaza defines the following entities and controllers to provide the above desc
 
 Entities:
 * [Cluster Environment](./docs/entities/clusterenvironment.md): represents an development environment on a kubernetes Cluster.
+* [Registered Service](./docs/entities/registeredservice.md): represents running instance of a software service.
 * Service Binding
 * Service Claim
 * Service Catalog


docs/entities/registeredservice.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Status
+
+The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). It contains only one required property to track state. The state could be one of the following: registered, available, claimed, unreachable. The registered service state is "registered" right after creation. If the health check is not defined or health check pass, the state would change to "available". Once a RegisteredService is claimed, its state moves to "claimed". On the other hand, if the health check comes back as a failure, the state would change to "unreachable".
+
+## Use Cases
+
+### Creation
+
+When a RegisteredService is created, Primaza will first check if health check can be run and modify the state accordingly. A ServiceCatalog is then updated to include the new RegisteredService if the state can be moved to "available".
```
sadlerap, line 34:
```suggestion
When a RegisteredService is created, Primaza will first check if the health check can be run and will modify the state accordingly. A ServiceCatalog will then updated to include the new RegisteredService if the state can be moved to "available".
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,45 @@
...
+by providing an image containing a client that can be run to test connectivity and authentication. This property is optional, when it is absent, it means the service will be considered available as soon as it is registered. 
+- SLA: Provides multiple levels of resiliency, scalability, fault tolerance and security. This allows claims to take into account the robustness of service. This property is optional, when it is absent, it means that there is no distinctions between services given the SLA.
+
+
+## Status
+
+The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). It contains only one required property to track state. The state could be one of the following: registered, available, claimed, unreachable. The registered service state is "registered" right after creation. If the health check is not defined or health check pass, the state would change to "available". Once a RegisteredService is claimed, its state moves to "claimed". On the other hand, if the health check comes back as a failure, the state would change to "unreachable".
```
sadlerap, line 28:
```suggestion
The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). It contains only one required property to track state. The state could be one of the following: registered, available, claimed, unreachable. The registered service state is "registered" right after creation. If the health check passes or is not defined, the state will change to "available". Once a RegisteredService is claimed, its state moves to "claimed". On the other hand, if the health check comes back as a failure, the state would change to "unreachable".
```

sadlerap, line 28: As far as I know, we also have `undetermined` as an available state.  We should call that out here and explain why it may be the state of a RegisteredService.

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,45 @@
+# RegisteredService
+
+A RegisteredService represents an instance of a running service. This resource can be created by both
+a human or a program. This flexibility allows for the service provisioning and service discovery operations
```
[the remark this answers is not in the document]
dperaza4dustbit: Do you know what I mean? I could be a human operator that enters the data, or a discovery program, we can call it agent maybe? Any suggestion to improve?

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,45 @@
...
+to be decoupled from representation of that service in a service catalog. The service catalog can be consumed by
+Kubernetes applications that are unaware of the lifecycle of the service. The creation of a registered server signals
+and update to the service catalog can be be advertised to the application developers. A RegisteredService can be
+claimed to be used by a specific application. If the application does not require the service any longer, it can
```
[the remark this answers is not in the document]
dperaza4dustbit: OK

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,45 @@
...
+### Deletion
+
+When a RegisteredService is deleted, the ServiceCatalog entry for the service should be deleted. Also, an claims for the service will change the state to pending since the matched registered service does not exist any longer.
+
+### Update
+
+When a RegisteredService is updated, a few things can happen:
+- Existing claims do not match the service any longer and they need to change to pending state
+- Existing service endpoint definition has change and should trigger a binding secret generation. 
+
```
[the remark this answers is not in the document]
dperaza4dustbit: OK

docs/entities/registeredservice.md
dperaza4dustbit: OK

docs/entities/registeredservice.md
dperaza4dustbit: OK

dperaza4dustbit: I know from the doc we defined that. But looking closely, when would the state be undetermined?

sadlerap: The way I understand it, `undetermined` can be useful when we don't actively know any of the other states.  It's a good way of indicating that something has gone wrong with the registered service, and that manual intervention may be required to resolve the issue.

As an example scenario for when this state would make sense, let's say a user defines a registered service with a health check that specifies a nonexistent container image.  The controller tries to run it, but it fails because the container image couldn't be pulled, resulting in a failure to run the health check.  In that case,  I think `undetermined` is the only state that makes sense.
 - `registered` to me means that primaza is aware of the resource, but doesn't know anything else about it.  It's effectively an acknowledgement that the resource exists.  However, in this case, we _do_ know more about it (namely, it has a faulty health check), so IMO `registered` doesn't convey the full context of what we've determined.  If we tried to run a health check that couldn't even run, we're well beyond the state of "we've acknowledged this resource exists".
 - `available` doesn't make sense to me here.  The user specified a health check, and so we should honor it by running it.  Since the health check couldn't determine availability, primaza shouldn't claim that we know otherwise.  IMO, primaza shouldn't hand out claims for services whose health check failed (regardless of _why_ it failed).
 - `claimed` makes no sense in this scenario, since there isn't a claim for the service.  We'd need for it to progress to `available` first, and with a failing health check, that's not happening.
 - `unreachable` doesn't quite feel right to me, since the problem here isn't with the service itself (it could be alive and well for all we know), but rather with the health check we have.  I'd prefer to reserve `unreachable` for "we ran the health check, and it told us the service isn't available"; in this scenario, we have "we tried to run the health check, but that failed", which is semantically different.

sadlerap: An alternative idea is that we can stuff this underneath `unreachable`, but we provide more context to the user via a condition field in the status.  Either way, in my opinion we need tell the user that their service isn't available because of a problem in the registered service resource itself, not with the actual service.

dperaza4dustbit: Thanks for the explanation @sadlerap I'm with you here, your example is perfect way to visualize it. Adding this changed

docs/entities/registeredservice.md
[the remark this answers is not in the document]
dperaza4dustbit: I made some changes here to try to make clear what it means. Let me know @sadlerap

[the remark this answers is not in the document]
sadlerap: Looks good

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,49 @@
+# RegisteredService
+
+A RegisteredService represents an instance of a running service. This resource can be created by both
+a human operator or a piece of software (agent, bot, script). This flexibility allows for the service provisioning and service discovery operations
+to be decoupled from representation of that service in a service catalog. The service catalog can be consumed by
+Kubernetes applications that are unaware of the lifecycle of the service. The creation of a registered server signals
+and update to the service catalog can be be advertised to the application developers. A RegisteredService can be
```
filariow, line 7: This sentence is not clear to me. What do you mean here?
```
The creation of a registered server signals and update to the service catalog can be be advertised to the application developers.
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,49 @@
...
+The definition of a RegisteredService can be obtained directly from our [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). The specification contains two required properties:
+
+- ServiceClassIdentity: A set of key/value pairs that identify the service class. Examples of service class identity keys include type of service, and provider of service. This property is required.
+- ServiceEndpointDefinition: A set of key/value pairs that provides two pieces of information: the service connectivity and
+the service authentication. Values can be either a string or a reference to a secret field. This property is required.
+
+A RegisteredService also has three optional properties, which gives the user more control over the resource:
+
+- Constraints: Restrictions of the registered service. A constrain could be a specific environment where the
+service cannot be claimed from. As an example, a constrain could describe that I don't want this service to be used on any environment except for production. This property is optional, when it is absent, it means there is not constrains.
```
filariow, line 22:
```suggestion
- Constraints: Restrictions of the registered service. A constraint could be a specific environment where the
service cannot be claimed from. As an example, a constraint could describe that I don't want this service to be used on any environment except for production. This property is optional, when it is absent, it means there is not constraints.
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+## Status
+
+The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). It contains only one required property to track state. The state could be one of the following: registered, available, claimed, unreachable or undetermined. The registered service state is "registered" right after creation. In most cases this will change immediately after the controller gets notified of existence of new resource, unless controller fails even before it tries to run health check. If the health check passes or is not defined, the state will change to "available". Once a RegisteredService is claimed, its state moves to "claimed". On the other hand, if the health check comes back as a failure, the state would change to "unreachable". In some cases, the health check could failed before it can determine where it fails of passes. In those cases, the state of the registered service will go to "undetermined" indicating there is an issue with the health check not with the service itself.
```
filariow, line 30:
```suggestion
The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml). It contains only one required property to track state. The state could be one of the following: registered, available, claimed, unreachable or undetermined. The registered service state is "registered" right after creation. In most cases this will change immediately after the controller gets notified of existence of new resource, unless controller fails even before it tries to run health check. If the health check passes or is not defined, the state will change to "available". Once a RegisteredService is claimed, its state moves to "claimed". On the other hand, if the health check comes back as a failure, the state would change to "unreachable". In some cases, the health check could fail executing, even before it can determine whether the service is healthy or not. In those cases, the state of the registered service will go to "undetermined" indicating there is an issue with the health check not with the service itself.
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,49 @@
...
+
+There are some situations when health check will fail while the registered service is in "claimed" state. In those cases the registered service will move to "unreachable". If, at a later time, the health check passes then the controller will check if there is still a claim matching the registered service and move the state back to "claimed". However, if there is not claim matching the registered service the state will move to "available"
```
filariow, line 32:
```suggestion
There are some situations when health checks will fail while the registered service is in "claimed" state. In those cases the registered service will move to "unreachable". If, at a later time, the health check passes then the controller will check if there is still a claim matching the registered service and move the state back to "claimed". However, if there is not claim matching the registered service the state will move to "available"
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Use Cases
+
+### Creation
+
+When a RegisteredService is created, Primaza will first check if the health check can be run and will modify the state accordingly. The ServiceCatalog will be updated to include the new RegisteredService only if the state can be moved to "available".
+
+### Deletion
+
+When a RegisteredService resource  is deleted, the ServiceCatalog entry for the service should be deleted. Also, an claims for the service will change the state to pending since the matched registered service does not exist any longer. Also, when a RegisteredService resource state changes to "claimed" the corresponding entry in the ServiceCatalog resource will be removed.
```
filariow, line 42:
```suggestion
When a RegisteredService resource  is deleted, the ServiceCatalog entry for the service should be deleted. Also, if the service is claimed, the ServiceClaim for the service will change the state to pending since the matched registered service does not exist any longer. Also, when a RegisteredService resource state changes to "claimed" the corresponding entry in the ServiceCatalog resource is removed.
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Update
+
+When a RegisteredService is updated, a few things can happen:
+- Existing claims do not match the service any longer and they need to change to pending state.
```
filariow, line 47: for each RegisteredService just one Claim can exists
```suggestion
- If claimed, the claim does not match the service any longer and it needs to change to pending state.
```

filariow (changes requested):
a few comments

docs/entities/registeredservice.md
dperaza4dustbit: changed

docs/entities/registeredservice.md
dperaza4dustbit: done

docs/entities/registeredservice.md
dperaza4dustbit: changed to "the health check"

docs/entities/registeredservice.md
dperaza4dustbit: done

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,68 @@
...
+
+The Status of the Service it is also defined under the [RegisteredService CRD](../../config/crd/bases/primaza.io_registeredservices.yaml).
+It contains only one required property to track state.
+The state could be one of the following: registered, available, claimed, unreachable or undetermined.
+The registered service state is "registered" right after creation.
+In most cases this will change immediately after the controller gets notified of existence of new resource, unless controller fails even before it tries to run health check.
+If the health check passes or is not defined, the state will change to "available".
+Once a RegisteredService is claimed, its state moves to "claimed".
+On the other hand, if the health check comes back as a failure, the state would change to "unreachable".
+In some cases, the health check could failed before it can determine whether the service is healthy or not.
```
sadlerap, line 42:
```suggestion
In some cases, the health check could fail before it can determine whether the service is healthy or not.
```

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,68 @@
...
+## Use Cases
+
+### Creation
+
+When a RegisteredService is created, Primaza will first check if the health check can be run and will modify the state accordingly.
+The ServiceCatalog will be updated to include the new RegisteredService only if the state can be moved to "available".
+
+### Deletion
+
+When a RegisteredService resource  is deleted, the ServiceCatalog entry for the service should be deleted.
```
sadlerap, line 59: Nit: extra space

```suggestion
When a RegisteredService resource is deleted, the ServiceCatalog entry for the service should be deleted.
```

filariow (changes requested):
a few more nits

docs/entities/registeredservice.md
dperaza4dustbit: OK, thanks

docs/entities/registeredservice.md
dperaza4dustbit: got it

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,68 @@
+# RegisteredService
+
+A RegisteredService represents an instance of a running service.
+This resource can be created by both a human operator or a piece of software (agent, bot, script). This flexibility allows for the service provisioning and service discovery operations to be decoupled from representation of that service in a service catalog.
+The service catalog can be consumed by Kubernetes applications that are unaware of the lifecycle of the service.
+The creation of a registered service signals and update to the service catalog.
```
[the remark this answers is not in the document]
dperaza4dustbit: thanks

docs/entities/registeredservice.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Creation
+
+When a RegisteredService is created, Primaza will first check if the health check can be run and will modify the state accordingly.
+The ServiceCatalog will be updated to include the new RegisteredService only if the state can be moved to "available".
+
+### Deletion
+
+When a RegisteredService resource  is deleted, the ServiceCatalog entry for the service should be deleted.
+Also, if a register service is claimed, the ServiceClaim resource state will be changed to "pending" by the service claim controller since the matched registered service does not exist any longer.
```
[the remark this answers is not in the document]
dperaza4dustbit: OK

filariow (approved):
lgtm

sadlerap (approved):

filariow (approved):
lgtm

PR #42, opened 2023-02-21
Publish images to Github Container Registry

filariow:
Example of action logs: https://github.com/filariow/primaza/actions/runs/4232055616/jobs/7351325676
Example of published packages: https://github.com/filariow?tab=packages&repo_name=primaza

### DIFF
--- a/.github/workflows/push-ghcr.yaml
+++ b/.github/workflows/push-ghcr.yaml
@@ -1,0 +1,66 @@
+name: Create and publish a Docker image
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches: ['main']
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/push-ghcr.yaml
```diff
@@ -0,0 +1,52 @@
...
+        env:
+          IMG_BASE: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-agentapp
+
+      - name: Build and push AgentSvc Docker image
+        run: |
+            IMG=$IMG_BASE:$(git rev-parse --short HEAD) make agentsvc docker-build docker-push
+            docker tag "$IMG_BASE:$(git rev-parse --short HEAD)" "$IMG_BASE:latest"
+            docker push "$IMG_BASE:latest"
+        env:
+          IMG_BASE: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}-agentsvc
```
sadlerap, line 52: We should push all images at once: if an image fails to build, none of the other images for that commit to be pushed into the image repository.

dperaza4dustbit (commented):
Where are we planning to publish this, I say lets create a new organization in quay

filariow:
> Where are we planning to publish this, I say lets create a new organization in quay

using this workflow as it is, we will publish the images in GitHub, and will find them in the repo's packages.
Take a look at https://github.com/filariow/primaza, on the right you find the section Package.

I propose to use this for the moment, and if we decide to move to quay we will just need to change the environment variables for repository, username and password. What do you think?

sadlerap:
`ghcr.io` as a container registry works for me -- I'm using it on another project, and it's a little easier to get going than quay is (since it's external to github).  Long-term, there's nothing stopping us from hosting containers in multiple repositories.

sadlerap (approved):
lgtm

PR #49, opened 2023-02-23
Parallel acceptance tests in Github Actions

filariow:


Kartikey-star (approved):
LGTM

sadlerap (dismissed):
I like this approach better than #25, so this should supercede that approach.

LGTM!

sadlerap (approved):
I don't know if github will pick up on this automatically, but merging this should close #25.

LGTM!

PR #56, opened 2023-03-03
Forward Service Classes to matching service namespaces

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #61, opened 2023-03-03
Delete Service Class

filariow:
Depends on "Forward Service Classes to service namespaces (#56)"


controllers/clusterenvironment_controller.go
```diff
@@ -60,6 +60,7 @@ const (
 type ClusterEnvironmentReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
+	Mapper meta.RESTMapper
```
Kartikey-star, line 63: Shouldn't we be using the mapper from the client itself?

Kartikey-star (changes requested):
As the restMapper is not being initialize i believe we can make of that interface available from the controller client itself.

Kartikey-star (approved):

PR #62, opened 2023-03-05
Update RegisteredService state when Claims are successful

dperaza4dustbit:
When a service claim is successful the matched regitered service state should be changed to "Claimed". This will make sure the service catalog does not contain the claimed service.

### DIFF
--- a/controllers/serviceclaim_controller.go
+++ b/controllers/serviceclaim_controller.go
@@ -35,6 +35,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/primaza/primaza/api/v1alpha1"
 	primazaiov1alpha1 "github.com/primaza/primaza/api/v1alpha1"
+	"github.com/primaza/primaza/pkg/envtag"
 	"github.com/primaza/primaza/pkg/slices"
 )
 
@@ -80,18 +81,12 @@ func (r *ServiceClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	rsName, err := r.processServiceClaim(ctx, req, rsl, sclaim)
+	err := r.processServiceClaim(ctx, req, rsl, sclaim)
 	if err != nil {
-		l.Error(err, "unable while processing ServiceClaim")
+		l.Error(err, "unable to process ServiceClaim")
 		return ctrl.Result{}, err
 	}
-	sclaim.Status.State = "Resolved"
-	sclaim.Status.ClaimID = uuid.New().String()
-	sclaim.Status.RegisteredService = rsName
-	if err := r.Status().Update(ctx, &sclaim); err != nil {
-		l.Error(err, "unable to update the ServiceClaim", "ServiceClaim", sclaim)
-		return ctrl.Result{}, err
-	}
+
 	return ctrl.Result{}, nil
 }
 
@@ -115,12 +110,13 @@ func checkSCISubset(serviceClaim, registeredService []v1alpha1.ServiceClassIdent
 	return true
 }
 
+
 func (r *ServiceClaimReconciler) extractServiceEndpointDefinition(
 	ctx context.Context,
 	req ctrl.Request,
 	rs v1alpha1.RegisteredService,
 	SEDKeys []string,
-	secret *corev1.Secret) (string, int, error) {
+	secret *corev1.Secret) (int, error) {
 	l := log.FromContext(ctx)
 	count := 0
 	// loop over the ServiceEndpointDefinition array part of RegisteredService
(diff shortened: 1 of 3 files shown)


test/acceptance/features/serviceClaim.feature
```diff
@@ -74,6 +74,8 @@ Feature: Service claim with label selector
         """
         Then On Primaza Cluster "primaza-main", the status of ServiceClaim "sc-test" is "Resolved"
         And On Primaza Cluster, "primaza-main", the secret "sc-test" has the key "type" with value "psqlserver"
+        And On Primaza Cluster "primaza-main", RegisteredService "primaza-rsdb" state will eventually move to "Claimed"
+        And On Primaza Cluster "primaza-main", ServiceCatalog "primaza-service-catalog" will not contain RegisteredService "primaza-rsdb"
```
filariow, line 78: we could add here also a test for the scenario in which SED is not matching

controllers/serviceclaim_controller.go
```diff
@@ -80,18 +80,26 @@ func (r *ServiceClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-	sclaim.Status.RegisteredService = rsName
+	sclaim.Status.RegisteredService = rs.Name
 	if err := r.Status().Update(ctx, &sclaim); err != nil {
 		l.Error(err, "unable to update the ServiceClaim", "ServiceClaim", sclaim)
 		return ctrl.Result{}, err
 	}
+
+	// Update RegisteredService status
+	rs.Status.State = primazaiov1alpha1.RegisteredServiceStateClaimed
+	if err := r.Status().Update(ctx, &rs); err != nil {
```
filariow, line 98: if we fail here, the registered service may be claimed from another ServiceClaim. I'd update the RegisteredService  status before the ServiceClaim's one.
If we proceed this way, we should also check if in a previous reconciliation an error occurred while updating ServiceClaim's status. This could be done by adding a property into the RegisteredService's status referencing to the binded Claim (e.g., `ClaimRef`). When reconciling the service class we should check if exists a RegisteredService with this field populated to the current ServiceClaim. In this case we update the ServiceClaim status again.

filariow (changes requested):

controllers/serviceclaim_controller.go
```diff
@@ -177,15 +187,16 @@ func (r *ServiceClaimReconciler) processServiceClaim(
 		// ServiceClassIdentity given in the RegisteredService
 		if checkSCISubset(sclaim.Spec.ServiceClassIdentity, rs.Spec.ServiceClassIdentity) {
 			registeredServiceFound = true
+			registeredService = rs
 			// loop over the environments given in the constraints part of the RegisteredService
 			for _, e := range rs.Spec.Constraints.Environments {
-				// check if the EnvironmentTag of ServiceClaim matches tat of the environment constraint
+				// check if the EnvironmentTag of ServiceClaim matches that of the environment constraint
 				if sclaim.Spec.EnvironmentTag == e {
```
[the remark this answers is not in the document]
dperaza4dustbit: This is getting more into the logic that @baijum added, but I will give it a try. If I get stuck we can always add another story to refactor.

test/acceptance/features/serviceClaim.feature
dperaza4dustbit: Agree, will give it a try, but same comment, would expect @baijum to pick up if I get stuck

controllers/serviceclaim_controller.go
```diff
@@ -115,12 +109,26 @@ func checkSCISubset(serviceClaim, registeredService []v1alpha1.ServiceClassIdent
...
+	}
+
+	for _, env := range constraintEnvs {
+		if envTag == env {
+			return true
+		}
+	}
+	return false
+}
+
```
filariow, line 125: this can be removed in favor of `envtag.Match`


 [https://github.com/primaza/primaza/blob/47502e5b8f41cfa49648ee153ecbe9fbb39194fe/pkg/envtag/envtag.go#L32](https://github.com/primaza/primaza/blob/47502e5b8f41cfa49648ee153ecbe9fbb39194fe/pkg/envtag/envtag.go#L32)

controllers/serviceclaim_controller.go
```diff
@@ -167,27 +186,25 @@ func (r *ServiceClaimReconciler) processServiceClaim(
...
 	count := 0
 	// loop over every RegisteredService
 	registeredServiceFound := false
+	var registeredService primazaiov1alpha1.RegisteredService
 	for _, rs := range rsl.Items {
 		// Check if the ServiceClassIdentity given in ServiceClaim is a subset of
 		// ServiceClassIdentity given in the RegisteredService
-		if checkSCISubset(sclaim.Spec.ServiceClassIdentity, rs.Spec.ServiceClassIdentity) {
+		if checkSCISubset(sclaim.Spec.ServiceClassIdentity, rs.Spec.ServiceClassIdentity) &&
+			environmentConstraintMet(rs.Spec.Constraints.Environments, sclaim.Spec.EnvironmentTag) {
```
filariow, line 199:
```suggestion
			envtag.Match(sclaim.Spec.EnvironmentTag, rs.Spec.Constraints.Environments) {
```

filariow (changes requested):

PR #81, opened 2023-03-21
Fix permissions with periodic security checks

sadlerap:
The actions job needs the ability to write security events, which it doesn't get automatically.  In these cases, we need to give the job explicit permission to do so.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -1,5 +1,9 @@
 name: Security checks
 
+permissions:
+  contents: read
+  security-events: write
+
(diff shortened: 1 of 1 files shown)


sadlerap:
Push events look good in my fork:
- [gosec](https://github.com/sadlerap/primaza/actions/runs/4480972758/jobs/7877008711)
- [trivy](https://github.com/sadlerap/primaza/actions/runs/4480972758/jobs/7877008377)

filariow (approved):
lgtm

PR #82, opened 2023-03-22
fix operator-sdk scaffolding

sadlerap:
When we invoke operator-sdk to introduce new project scaffolding, we run into the issue that it generates resources under the `primaza.io.primaza.io` group name, not `primaza.io`.  This is caused by a misconfiguration in the PROJECT file: we set both the domain and the group field for each resource, which is not correct in this circumstance.

To prevent problems from occuring in the future, remove the group field.

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -12,7 +12,6 @@ resources:
     namespaced: true
   controller: true
   domain: primaza.io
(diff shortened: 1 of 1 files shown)


filariow (approved):
lgtm

PR #83, opened 2023-03-23
Validate Service Claim fields

baijum:
Jira: https://issues.redhat.com/browse/APPSVC-1298

### DIFF
--- a/api/v1alpha1/serviceclaim_webhook.go
+++ b/api/v1alpha1/serviceclaim_webhook.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2023 The Primaza Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// log is for logging in this package.
+var serviceclaimlog = logf.Log.WithName("serviceclaim-resource")
+
+type serviceClaimValidator struct {
+	client client.Client
+}
+
+func (r *ServiceClaim) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		WithValidator(&serviceClaimValidator{
+			client: mgr.GetClient(),
+		}).
+		Complete()
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-primaza-io-v1alpha1-serviceclaim,mutating=false,failurePolicy=fail,sideEffects=None,groups=primaza.io,resources=serviceclaims,verbs=create;update,versions=v1alpha1,name=vserviceclaim.kb.io,admissionReviewVersions=v1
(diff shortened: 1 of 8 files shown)


test/acceptance/features/serviceClaim.feature
```diff
@@ -747,3 +747,83 @@ Feature: Service claim with label selector
         And  On Worker Cluster "worker", Service Binding "sc-test" does not exist in "applications"
         And  On Worker Cluster "worker", the secret "sc-test" does not exist in namespace "applications"
 
+    Scenario: Create ServiceClaim with ApplicationClusterContext and EnvironmentTag
```
filariow, line 750: I'd propose to implement this test in `serviceclaim_webhook_test.go`. WDYT?

api/v1alpha1/serviceclaim_webhook.go
```diff
@@ -0,0 +1,98 @@
...
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (v *serviceClaimValidator) ValidateUpdate(ctx context.Context, oldObj runtime.Object, newObj runtime.Object) error {
+	r, ok := newObj.(*ServiceClaim)
+	if !ok {
+		err := fmt.Errorf("Object is not a Service Claim")
+		serviceclasslog.Error(err, "Attempted to validate non-ServiceClaim resource", "gvk", newObj.GetObjectKind().GroupVersionKind())
+		return err
+	}
+
```
filariow, line 80: I'd propose to check for immutability of the following fields:
- Application
- EnvironmentTag
- ApplicationClusterContext

WDYT?

filariow (changes requested):
some nits

test/acceptance/features/serviceClaim.feature
baijum: @filariow 
There is a test : https://github.com/primaza/primaza/pull/83/files#diff-e412822a13ddd16751410c7c39a421605cca061013c078baf4fa953e7e8cdfc0R40
Please check!

api/v1alpha1/serviceclaim_webhook.go
baijum: How is that immutability works? So, the original values cannot be modified?

test/acceptance/features/serviceClaim.feature
filariow: right, my concern here is that this test does not have the 3 steps Given-When-Then and that's validating the ServiceClaim's yaml. We may have this kind of validation just in `serviceclaim_webhook_test.go` without duplicating it. Or may it be a good reason to check it also here?

api/v1alpha1/serviceclaim_webhook.go
filariow: yes. To me a Claim should be deleted and created again if some of those field changes. Otherwise reconciliation could be very hard to do. Would be hard to track ServiceBindings created before changing for example the EnvironmentTag

test/acceptance/features/serviceClaim.feature
baijum: The test in the `serviceclaim_webhook_test.go` is a unit test. So it has the minimum required test fixture for the test. What we are writing using these feature files is functional/acceptance tests. These are complementary.

api/v1alpha1/serviceclaim_webhook.go
baijum: Can we create a separate user story and implement it?

filariow: yes that is a good idea

test/acceptance/features/serviceClaim.feature
filariow: ok

api/v1alpha1/serviceclaim_webhook.go
```diff
@@ -0,0 +1,98 @@
...
+	serviceclaimlog.Info("validate create", "name", r.Name)
+	return v.validate(r)
+}
+
+func (v *serviceClaimValidator) validate(r *ServiceClaim) error {
+	if r.Spec.ApplicationClusterContext != nil && r.Spec.EnvironmentTag != "" {
+		return fmt.Errorf("Both ApplicationClusterContext and EnvironmentTag cannot be used together")
+	}
```
sadlerap, line 67: do we also need to enforce that at least one of them is set?

sadlerap (commented):
One question, otherwise looks good.

api/v1alpha1/serviceclaim_webhook.go
baijum: @sadlerap Thanks! I have implemented it.

filariow (approved):
lgtm

PR #87, opened 2023-03-28
On ServiceClaim deletion, remove ServiceBinding and Secret

Kartikey-star:


pkg/primaza/controlplane/application_namespaces.go
```diff
@@ -127,3 +128,36 @@ func PushServiceCatalogToApplicationNamespaces(ctx context.Context, sc primazaio
...
+	var errs []error
+	for _, ns := range namespaces {
+		sb := &primazaiov1alpha1.ServiceBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      sc.Name,
+				Namespace: ns,
+			},
+		}
+
+		secret := &v1.Secret{
```
filariow, line 142:
```suggestion
		secret := &corev1.Secret{
```

controllers/serviceclaim_controller.go
```diff
@@ -343,6 +411,55 @@ func (r *ServiceClaimReconciler) pushToClusterEnvironments(
...
+		ce, err := r.getEnvironmentFromClusterEnvironment(ctx, req, sclaim.Spec.ApplicationClusterContext.ClusterEnvironmentName)
+		if err != nil {
+			l.Info("error getting ClusterEnvironment", "error", err)
+			return err
+		}
+		cli, err := clustercontext.CreateClient(ctx, r.Client, *ce, r.Scheme, r.Client.RESTMapper())
+		if err != nil {
+			return err
+		}
+		if err = controlplane.DeleteServiceBindingAndSecretFromNamespaces(ctx, cli, sclaim, ce.Spec.ApplicationNamespaces); err != nil {
```
filariow, line 433: The ApplicationClusterContext defines also the target namespace

```suggestion
                ns := []string{sclaim.Spec.ApplicationClusterContext.Namespace}
		if err = controlplane.DeleteServiceBindingAndSecretFromNamespaces(ctx, cli, sclaim, ns); err != nil {
```

controllers/serviceclaim_controller.go
```diff
@@ -71,6 +74,33 @@ func (r *ServiceClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	l.Info("Check If service binding is deleted")
```
filariow, line 77:
```suggestion
	l.Info("Check if Service Claim is marked for deletion")
```

pkg/primaza/controlplane/application_namespaces.go
```diff
@@ -127,3 +128,36 @@ func PushServiceCatalogToApplicationNamespaces(ctx context.Context, sc primazaio
...
+			if !apierrors.IsNotFound(err) {
+				errs = append(errs, err)
+			}
+
+		}
+		if err := cli.Delete(ctx, secret, &client.DeleteOptions{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				errs = append(errs, err)
+			}
+		}
```
filariow, line 159: another approach may be to leverage on Ownership for this. It would reduce the probability of having an error and having to reconcile again. We should modify the `controlplane.pushServiceBindingToNamespace` to push the ServiceBinding before creating the Secret and enriching the Secret with Ownership details from ServiceBinding.

PTAL: https://github.com/filariow/kid/blob/4e10389547c1bd78e0526cb80a8d2dd8c5671681/pkg/kube/secret.go#L72

After applied this change we should just delete the ServiceBinding. This would also remove the requirement for the `delete` permission added in `kubernetesapi.py`.

WDYT?

filariow (changes requested):
some comments

pkg/primaza/controlplane/application_namespaces.go
Kartikey-star: Yeah i believe, that is better as we won't need to explicitly set the permission to delete secret. I have made the change in pushservicebinding method to create secret with owner reference set to service binding uid.

controllers/clusterenvironment_controller.go
```diff
@@ -63,7 +63,7 @@ type ClusterEnvironmentReconciler struct {
 	Scheme *runtime.Scheme
 }
 
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
```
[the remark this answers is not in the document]
Kartikey-star: Yeah makes sense

filariow (approved):
lgtm

PR #90, opened 2023-03-29
Minor fixes to tests and manifests

filariow:


make/agents/svc/deploy.mk
```diff
@@ -14,6 +14,10 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build $(KUSTOMIZE_ARGS) $(KUSTOMIZE_OVERLAY) | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
+.PHONY: deploy-rbac
+deploy-rbac:
+	$(KUSTOMIZE) build config/agents/svc/rbac | kubectl apply -n $(NAMESPACE) -f -
+
```
sadlerap, line 20: I'd like to keep operator installation as unified as possible.  Why are we splitting rbac rules out?  Is there any reason why `make agentsvc deploy` doesn't work in this case?

filariow: the deployment should be pushed from Primaza's ClusterEnvironment controller when the namespace is added to the `ClusterEnvironment.serviceNamespaces`. The deployment of RBAC will be eventually implemented in the `create namespace` command of `primazactl`.

config/agents/svc/rbac/kustomization.yaml
```diff
@@ -6,3 +6,4 @@ resources:
 - service_agent_manager_role_binding.yaml
 - service_agent_manager_role.yaml
 - service_agent_account.yaml
+namespace: services
```
sadlerap, line 9: Since `system` is used everywhere else, shouldn't we be using that as the default for `kustomize`?

When we do deployments in acceptance testing, we can always override it either by changing this namespace with `kustomize edit` or by overriding it in `kubectl`.

test/acceptance/features/appNsServiceBinding.feature
sadlerap, line 1: rename to `appNamespaceServiceBinding.feature`

test/acceptance/features/appNsServiceBinding.feature
```diff
@@ -0,0 +1,341 @@
...
+        """
+        apiVersion: primaza.io/v1alpha1
+        kind: ClusterEnvironment
+        metadata:
+            name: worker
+            namespace: primaza-system
+        spec:
+            environmentName: dev
+            clusterContextSecret: primaza-kw
+        """
```
sadlerap, line 19: A lot of these look like they can be moved to background to avoid duplication.

sadlerap (changes requested):
a few nits

config/agents/svc/rbac/kustomization.yaml
filariow: nice catch, this change was not supposed to be commited

test/acceptance/features/appNsServiceBinding.feature
filariow: too long, kind was breaking

config/agents/svc/kustomization.yaml
```diff
@@ -51,4 +51,3 @@ images:
   newTag: latest
 patches:
 - path: ../../default/webhookcainjection_patch.yaml
-namespace: services
```
sadlerap, line 54: We should probably keep this in, as a mechanism to autoreplace namespaces in the manifest we generate.

filariow: good idea! I applied the same suggestion to Application Agent's manifests

sadlerap (approved):

PR #92, opened 2023-03-30
Update manifests to reduce required permissions

filariow:
depends on #90

### DIFF
--- a/config/rbac/role_binding.yaml
+++ b/config/rbac/role_binding.yaml
@@ -1,5 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
   labels:
     app.kubernetes.io/name: clusterrolebinding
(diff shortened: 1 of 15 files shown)


config/rbac/role_binding.yaml
```diff
@@ -9,11 +9,11 @@ metadata:
     app.kubernetes.io/part-of: primaza
     app.kubernetes.io/managed-by: kustomize
   name: manager-rolebinding
+  namespace: system
```
dperaza4dustbit, line 12: Where does system come from is this supposed to be a variable with the namespace for this primaza tenant?

filariow: That's just a placeholder that can be used in kustomize manifests to identify objects. Kustomize's default seemed to be `system` so we are relying on this value. We can choose a different value if we prefer.

Default usage of these manifests would rely on overwriting this value.

dperaza4dustbit (approved):
lgtm

dperaza4dustbit (approved):
lgtm

PR #94, opened 2023-03-30
Fix Make 3 compatibility problem

filariow:
Signed-off-by: Francesco <fr.ilario@gmail.com>


Kartikey-star (approved):
Works fine with make 3.8

PR #95, opened 2023-03-30
delete registered services on service class deletion

sadlerap:
This adds a finalizer to service classes on service agents.  When deleted, we delete the registered services that this service class generated.

### DIFF
--- a/controllers/agents/svc/serviceclass_controller.go
+++ b/controllers/agents/svc/serviceclass_controller.go
@@ -24,6 +24,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -55,6 +56,8 @@ type ServiceClassReconciler struct {
 	Mapper       meta.RESTMapper
 }
 
+const finalizer = "serviceclasses.primaza.io/finalizer"
+
 func NewServiceClassReconciler(mgr ctrl.Manager) *ServiceClassReconciler {
 	return &ServiceClassReconciler{
 		Client:    mgr.GetClient(),
@@ -79,8 +82,19 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	serviceClass := v1alpha1.ServiceClass{}
 	err := r.Get(ctx, req.NamespacedName, &serviceClass)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// service class was deleted on the control plane; we've got
+			// nothing more to do here.
+			return ctrl.Result{}, nil
+		}
+		// something went wrong, requeue
 		reconcileLog.Error(err, "Failed to retrieve ServiceClass")
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, err
+	}
+
+	if err = r.setOwnerReference(ctx, &serviceClass, req.Namespace); err != nil {
+		reconcileLog.Error(err, "Failed to set owner reference on ServiceClass", "namespace", req.Namespace, "name", req.Name)
+		return ctrl.Result{}, err
 	}
 
 	// next, get all the services that this service class controls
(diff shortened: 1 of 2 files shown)


controllers/agents/svc/serviceclass_controller.go
```diff
@@ -93,7 +102,45 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			}
+			return err
+		}
+
+		// remove the finalizer so we don't requeue
+		if controllerutil.RemoveFinalizer(&serviceClass, finalizer) {
+			if err = r.Update(ctx, &serviceClass, &client.UpdateOptions{}); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
filariow, line 138: as the handler is executed at L143, the finalizer will be removed even if handler fails. Also deletion handler func should check if it can ignore the error (i.e. NotFoundError) or not. If not, e.g. network error, we should retry before deleting the finalizer.

Would it be a solution to move this logic into the handler?

filariow (changes requested):
some nits

controllers/agents/svc/serviceclass_controller.go
sadlerap: I've changed this logic, PTAL.

filariow (dismissed):
a few optional nits

controllers/agents/svc/serviceclass_controller.go
filariow: Logic lgtm, but code needs formatting

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -95,10 +104,57 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	// then, write all the registered services up to the primaza cluster
-	err = r.CreateRegisteredServices(ctx, &serviceClass, *services)
-	if err != nil {
-		reconcileLog.Error(err, "Failed to write registered services")
-		// fallthrough: we still want to write the service class status field
+	if serviceClass.DeletionTimestamp.IsZero() {
```
Kartikey-star, line 107: This conditional block is preventing the control to reach set ownership method. I believe we need to place that function above these checks so as to resolve the errors on service namespace cleanup.

sadlerap: I tested this locally, this doesn't have an impact on failing tests.  Which makes sense: if we've hit the deletion case, we've already created the service class, and so we've already set the owner reference on it.

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -151,11 +209,18 @@ func (r *ServiceClassReconciler) CreateRegisteredServices(ctx context.Context, s
...
 	serviceClass.Status.Conditions = append(serviceClass.Status.Conditions, metav1.Condition{
-		Type:    "Connection",
-		Message: status.Message,
-		Reason:  string(status.Reason),
-		Status:  metav1.ConditionStatus(status.State),
+		Type:               "Connection",
+		Message:            status.Message,
+		Reason:             string(status.Reason),
+		Status:             metav1.ConditionStatus(state),
+		LastTransitionTime: metav1.Now(),
```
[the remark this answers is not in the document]
sadlerap: Yes, fixed.

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -151,11 +209,18 @@ func (r *ServiceClassReconciler) CreateRegisteredServices(ctx context.Context, s
 	// TODO(sadlerap): move TestConnection from `workercluster` into a more
 	// general-purpose package
 	status := workercluster.TestConnection(ctx, config)
+	state := "Unknown"
+	if status.State == v1alpha1.ClusterEnvironmentStateOnline {
+		state = "True"
+	} else if status.State == v1alpha1.ClusterEnvironmentStateOffline {
+		state = "False"
+	}
```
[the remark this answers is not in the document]
sadlerap: Good catch - fixed.

controllers/agents/svc/serviceclass_controller.go
sadlerap: should be fixed now

controllers/agents/svc/serviceclass_controller.go
sadlerap: Yep, CI fails; check the [CI output](https://github.com/primaza/primaza/actions/runs/4601246839/jobs/8128901937?pr=95) for commit 896c289.

Kartikey-star: Yeah it makes sense. Are you able to replicate the test failure locally? Test are passing locally for me. Created a PR https://github.com/primaza/primaza/pull/101 to use scenario-id instead of static name for service class and increasing timeout to 90s , but still see it failing on ci.

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -95,15 +109,57 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			if err = r.Update(ctx, &serviceClass, &client.UpdateOptions{}); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
+		// nothing to do
+		return ctrl.Result{}, nil
 	}
 
 	// finally, write the status of the service class
```
Kartikey-star, line 165: This part of updating the status of service class is unreachable, so i guess we can remove the else block and proceed with status update or return nil error

sadlerap:
I think we're getting failures in `serviceNamespaceCleanup.feature` because we're now setting a finalizer on service class resources, which is triggering this behavior.

As far as I can tell, the controller sometimes get a deletion request, but it usually doesn't fully process it because the pod is killed before the deletion request can be completed.

This is happening because we set an owner reference on the deployment, not the pod underlying the controller.  Because of this, kubernetes doesn't identify the dependency between service classes and the controller correctly; it thinks that killing the pod before all the service classes have been finalized is entirely valid.  Setting the pod as an owner reference isn't valid here: if the ReplicaSet controller decides during normal operation to delete the pod we set an owner reference on, then all of the service classes are going to be deleted.

If we want to support service class deletion when cluster environments change (and I think we should!), we're going to need to have the cluster environment controller delete the service classes before deleting the deployment.

filariow:
> I think we're getting failures in `serviceNamespaceCleanup.feature` because we're now setting a finalizer on service class resources, which is triggering this behavior.

I completely agree with @sadlerap's analysis. It seems a [known problematic scenario with operators](https://github.com/operator-framework/operator-sdk/issues/762).

In alternative to deleting from Primaza's Control Plane which may require more privileges than we would like to have and much more load for the Control Plane, I'd propose implementing a finalizer on the agent's deployment and having the agent watch for events on its deployment. Once the agent's deployment is requested to be deleted, the controller would delete all Primaza resources and then remove the finalizer on the deployment itself. Eventually, it should also be implemented a mechanism that would disable the reconciliation of new Primaza resources if it's not just reconciling a deletion event.

What do you think?

Kartikey-star:
By adding a finalizer to agent svc deployment i am able to get a successful ci along with the commit for the current PR on https://github.com/primaza/primaza/pull/101

sadlerap:
Moving to draft until #101 is merged.

Kartikey-star (approved):
LGTM

sadlerap:
@Kartikey-star please re-approve, rebased.

Kartikey-star (approved):
LGTM

PR #97, opened 2023-03-31
run pr-checks on push to main

sadlerap:
This should help us determine when code landing in `main` isn't passing all checks.

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -4,6 +4,9 @@ on: # yamllint disable-line rule:truthy
   pull_request:
     branches:
       - main
+  push:
+    branches:
+      - main
 
 env:
   GO111MODULE: "on"


baijum (approved):

dperaza4dustbit (approved):
lgtm

filariow (approved):
lgtm

I was just wondering if we really need this as we are running this checks on PR and requiring a PR to contribute to `main`.

sadlerap:
@filariow I originally filed this PR since I thought the cause of the issues on #95 was a regression, and I wanted to have tooling to check `main` for regressions.  It turns out that wasn't the case for #95, but it would still be useful to have.

I've also noticed a few other projects throwing their ci checks on branch pushes (for instance, [argo-cd does it](https://github.com/argoproj/argo-cd/blob/master/.github/workflows/ci-build.yaml)).  As far as I can tell, it seems to be good practice to do this.

filariow (approved):
lgtm

PR #99, opened 2023-04-03
increase service namespace cleanup timeout

Kartikey-star:


test/acceptance/features/steps/kubernetesapi.py
```diff
@@ -276,8 +276,8 @@ def on_worker_cluster_check_service_class_not_exists_in_service_namespace(contex
                 plural="serviceclasses",
                 name=service_class),
             check_success=lambda x: x is not None,
-            step=1,
-            timeout=10)
+            step=5,
```
filariow, line 279: I'd propose to increase the check interval

```suggestion
            step=1,
```

filariow (changes requested):
proposal

sadlerap (approved):
lgtm, 60 seconds seems reasonable here.  If we still run into issues, we can always bump this up further.

PR #100, opened 2023-04-03
use correct image name in agentsvc and agentapp deployment

sadlerap:
This was causing image overrides not to work correctly during service agent deployments.

### DIFF
--- a/make/agents/svc/deploy.mk
+++ b/make/agents/svc/deploy.mk
@@ -22,7 +22,7 @@ deploy-rbac:
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/agents/svc && \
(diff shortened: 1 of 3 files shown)


make/agents/svc/deploy.mk
```diff
@@ -22,7 +22,7 @@ deploy-rbac:
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/agents/svc && \
-		$(KUSTOMIZE) edit set image controller=${IMG} && \
+		$(KUSTOMIZE) edit set image agentsvc=${IMG} && \
```
filariow, line 25: Nice catch! It seems to me we may need the same fix in Application Agent's makefiles and manifests.
Would it be a good idea to apply these changes in this PR or are we going to create a new one?

dperaza4dustbit (approved):
lgtm

filariow (approved):
lgtm

PR #101, opened 2023-04-04
Add Finalizer to Agent Service Deployment

Kartikey-star:


controllers/agents/svc/agentsvc_controller.go
```diff
@@ -0,0 +1,111 @@
...
+const agentsvcfinalizer = "agent.primaza.io/finalizer"
+
+func (r *AgentServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	l := log.FromContext(ctx)
+	l.Info("Reconcile Agent Service Deployment")
+
+	// first, get the agent svc deployment
+	agentsvcdeployment := appsv1.Deployment{}
+	objKey := client.ObjectKey{
+		Name:      "primaza-controller-agentsvc",
```
filariow, line 52: I'd suggest to move that in a `const`. A name may be `ServiceAgentDeploymentName`. Also, a great place where to have that const would be `pkg/primaza/constants/constants.go`

```suggestion
		Name:      constants.ServiceAgentDeploymentName,
```

controllers/agents/svc/agentsvc_controller.go
```diff
@@ -0,0 +1,111 @@
...
+		Namespace: req.Namespace,
+	}
+	err := r.Get(ctx, objKey, &agentsvcdeployment)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// service class was deleted on the control plane; we've got
+			// nothing more to do here.
+			return ctrl.Result{}, nil
+		}
```
filariow, line 61: this should never happen. If this happen, it means that the name of the deployment is different from `constants.ServiceAgentDeploymentName`. I'd propose to log the error and to crash if this happen.

```suggestion
		if apierrors.IsNotFound(err) {
		        // it should never happen that this controller does not find itself
		        log.Error(err,
		              "service agent deployment not found, that should be a bug",
		              "expected deployment name", constants.ServiceAgentDeploymentName)
			panic(err)
		}
```

filariow (changes requested):
a couple nits

controllers/agents/svc/agentsvc_controller.go
Kartikey-star: Yeah that's better

filariow (approved):
lgtm

PR #112, opened 2023-04-13
Implement SED secrets

sadlerap:
SED information could contain sensitive data, such as a password or authentication token.  Such data should not be stored within a registered service's resource itself.  When making registered services with a service class, we should give users the option to indicate that their data is secret.

This change introduces a `Secret` flag within a SED mapping, which indicates whether that particular mapping item should be stored within the registered service (false) or within an associated secret (true). By default, this flag is true, since security features such as this should be opt-out, not opt-in.

### DIFF
--- a/controllers/agents/svc/serviceclass_controller.go
+++ b/controllers/agents/svc/serviceclass_controller.go
@@ -48,6 +48,8 @@ import (
 // cluster, and `namespace`, the namespace to write registered services to
 const PRIMAZA_CONTROLLER_REFERENCE string = "primaza-kubeconfig"
 
+const finalizer = "serviceclasses.primaza.io/finalizer"
+
 // ServiceClassReconciler reconciles a ServiceClass object
 type ServiceClassReconciler struct {
 	client.Client
@@ -56,7 +58,11 @@ type ServiceClassReconciler struct {
 	Mapper       meta.RESTMapper
 }
 
-const finalizer = "serviceclasses.primaza.io/finalizer"
+type SEDMapping struct {
+	key    string
+	path   *jsonpath.JSONPath
+	secret bool
+}
 
 func NewServiceClassReconciler(mgr ctrl.Manager) *ServiceClassReconciler {
 	return &ServiceClassReconciler{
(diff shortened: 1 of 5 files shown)


controllers/agents/svc/serviceclass_controller.go
```diff
@@ -106,15 +112,24 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 				reconcileLog.Error(err, "Failed to create registered service",
 					"service", rs.Name,
 					"namespace", rs.Namespace)
 			} else {
 				reconcileLog.Info("Wrote registered service", "service", rs.Name, "namespace", rs.Namespace, "operation", op)
 			}
+
+			if secret != nil {
+				_, err1 := controllerutil.CreateOrUpdate(ctx, remote_client, secret, func() error { return controllerutil.SetOwnerReference(clone, secret, remote_client.Scheme()) })
+				err = errors.Join(err, err1)
```
Kartikey-star, line 131: We seem to be preparing the error List on L299, so we can avoid this operation here right?

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -130,15 +145,20 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 				if apierrors.IsNotFound(err) {
 					// we tried to delete an object that doesn't exist, so
 					return nil
 				}
 				reconcileLog.Error(err, "Failed to delete registered service", "namespace", rs.Namespace)
+				return err
 			}
-			return err
+
+			if secret != nil {
```
Kartikey-star, line 158: We are setting the Registered Service as owner of the secret in L130, so can deletion of secret be avoided here?

controllers/agents/svc/serviceclass_controller.go
sadlerap: I've changed the implementation to aggregate errors there.

controllers/agents/svc/serviceclass_controller.go
sadlerap: Yep, good catch.  Removed.

Kartikey-star (approved):
Looks good to me

PR #119, opened 2023-04-18
Remove duplicate serviceclaim role for app agent

mmulholla:
service claim role is was defined twice, once with delete and once without. 
Have removed the second one without delete as delete is required.


filariow (approved):

dperaza4dustbit (approved):
lgtm

PR #121, opened 2023-04-19
change agents to use published images

mmulholla:
Primazactl needs to use published images. It does not install the agent app or svc images, primaza does.  Therefore primaza should use published images by default. Acceptance tests changes them back to local images.


filariow (approved):
lgtm

PR #129, opened 2023-04-21
Move from CSR to SA in Primaza's Control Plane

filariow:
Move from Certificate Signing Requests (CSR) to Service Accounts (SA) for new Application and Service namespaces.

As of now, the CSR is created by `primazactl`.
As we are moving to SA and `primaza` is already authorized to create secrets on worker cluster, we can have `primaza` itself handle the life-cycle of the agent's Service Accounts.

This means that when a new Application or Service namespace is created, `primaza` create the Service Account, the Service Account Token (secret), bakes the kubeconfig and push it in the worker's namespace.
More precisely, ClusterEnvironment controller is in charge of pushing the kubeconfig secret to target namespaces.

### DIFF
--- a/config/manager/manager_configmap.yaml
+++ b/config/manager/manager_configmap.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: manager
+  namespace: system
+data:
+  cluster_host_external: ""
--- a/config/agents/svc/rbac/manager_role.yaml
+++ b/config/agents/svc/rbac/manager_role.yaml
@@ -41,3 +41,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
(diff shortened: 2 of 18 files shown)


config/manager/manager_configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: manager
+  namespace: system
+data:
+  cluster_host_external: ""
```
filariow, line 7: @mmulholla `primazactl`'s users should be able to set this value. As an example, in a two kind cluster environment, this should be set to the internal IP of the container that's running the kind cluster on which primaza is installed

filariow:
closed in favor of #132

filariow:
closed in favor of #132

PR #130, opened 2023-04-21
Svc Agent wait for Service Class deletion

filariow:


sadlerap:
Hmm, I'm not entirely convinced this is actually necessary.

When we trigger a deletion of the controller, every service class in this namespace is going to get deleted.  This is because we explicitly set an owner reference on them that points to the controller, so the delete request is going to cascade down to the service classes.

There aren't many service classes that aren't going to be deleted.  In particular, it's going to be service classes that were created after the controller's deletion request was filed.  Since the primaza controller won't be pushing new service classes to this namespace after it triggers the agent's deletion, that only leaves service classes created manually by the user within this namespace, which doesn't seem too likely to me (after all, they did trigger a deletion).

To be honest, if we really want to enforce this, we should make a [mutation webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) to always add the controller's ownership on admission, so that every service class gets its deletion request.  We also might want to use [foreground deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion) to make sure that service classes do get deleted before the controller does.

filariow:
@sadlerap We have to wait for service classes to be deleted because the agent is removing the finalizer on them. So we can not allow the agent to be terminated before the Service Classes are effectively deleted. Otherwise, we will left some service classes. It does not happen often in tests because we always use just one ServiceClass and most of the time the controller is able to remove the finalizer before being terminated.
Btw, I had cleanup tests failing for this reason and that change should fix it. 

We can reason on a better implementation for this mechanism, but I'd say that for now to fix flaky tests it may be merged.
Unfortunately, Mutating Webhooks are Cluster-scoped, and we don't want to have Cluster-scoped resources at this stage.

sadlerap:
Which tests were you seeing failures on?  I want to try something with webhooks, and I'd like a place to start.

filariow:
> Which tests were you seeing failures on? I want to try something with webhooks, and I'd like a place to start.

You can use this change to trigger the bug:

```
diff --git a/controllers/agents/svc/serviceclass_controller.go b/controllers/agents/svc/serviceclass_controller.go
index 13e89f5..c25acf1 100644
--- a/controllers/agents/svc/serviceclass_controller.go
+++ b/controllers/agents/svc/serviceclass_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 
@@ -170,6 +171,7 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 
 		// remove the finalizer so we don't requeue
+		time.Sleep(2 * time.Minute)
 		if controllerutil.RemoveFinalizer(&serviceClass, finalizer) {
 			if err = r.Update(ctx, &serviceClass, &client.UpdateOptions{}); err != nil {
 				return ctrl.Result{}, err
diff --git a/test/acceptance/features/serviceNamespaceCleanup.feature b/test/acceptance/features/serviceNamespaceCleanup.feature
index 2748572..7fe0108 100644
--- a/test/acceptance/features/serviceNamespaceCleanup.feature
+++ b/test/acceptance/features/serviceNamespaceCleanup.feature
@@ -50,6 +50,7 @@ Feature: Cleanup service namespace
             """
         And  On Worker Cluster "worker", Service Class "$scenario_id-serviceclass" exists in "services"
 
+    @wip
     Scenario: Service Class is removed on Service Namespace deletion
         When On Primaza Cluster "main", Resource is updated
             """
```

controllers/agents/svc/agentsvc_controller.go
```diff
@@ -70,6 +72,15 @@ func (r *AgentServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		if err := r.removeServiceClasses(ctx, req); err != nil {
 			return ctrl.Result{}, err
 		}
+
```
Kartikey-star, line 75: Not sure if the logic to add listing the service classes is required. We can simply reque if the err is not nil on L72. WDYT?

Kartikey-star (approved):

Kartikey-star (approved):

PR #131, opened 2023-04-21
Ignore error on ServiceClass already exists

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #135, opened 2023-04-24
update service classes when they already exist

sadlerap:


sadlerap:
Possibly replaces #131, what do you think @filariow?

filariow (approved):
lgtm!

PR #136, opened 2023-04-26
Update service binding and secret on application namespaces

Kartikey-star:
Ref: https://issues.redhat.com/browse/APPSVC-1341

### DIFF
--- a/api/v1alpha1/serviceclaim_types.go
+++ b/api/v1alpha1/serviceclaim_types.go
@@ -47,7 +47,7 @@ type ServiceClaimSpec struct {
 
 // ServiceClaimStatus defines the observed state of ServiceClaim
 type ServiceClaimStatus struct {
-	//+kubebuilder:validation:Enum=Pending;Resolved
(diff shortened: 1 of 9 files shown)


api/v1alpha1/serviceclaim_types.go
```diff
@@ -47,7 +47,7 @@ type ServiceClaimSpec struct {
 
 // ServiceClaimStatus defines the observed state of ServiceClaim
 type ServiceClaimStatus struct {
-	//+kubebuilder:validation:Enum=Pending;Resolved
+	//+kubebuilder:validation:Enum=Pending;Resolved;Invalid
```
baijum, line 50: If the webhook returns an error, the resource is not going to be created. Is that correct?
Then why do we need an `Invalid` state?

baijum: Please update the docs with this new state: https://github.com/primaza/primaza/blob/main/docs/entities/serviceclaim.md

baijum (approved):

filariow (approved):
lgtm

PR #137, opened 2023-04-26
Unconstrained ServiceClasses should apply to all environments

filariow:


controllers/serviceclass_controller.go
```diff
@@ -102,11 +102,6 @@ func (r *ServiceClassReconciler) reconcileEnvironments(ctx context.Context, sc *
 		return err
 	}
 
-	if sc.Spec.Constraints == nil {
```
baijum, line 105: Do we need a test case to ensure this behavior?

Kartikey-star (approved):

baijum (approved):

PR #138, opened 2023-04-27
[service-agent] disable normal reconciliation if controller is being deleted

sadlerap:
If the controller has had its `deletionTimestamp` set, we shouldn't try to reconcile it.

### DIFF
--- a/controllers/agents/svc/serviceclass_controller.go
+++ b/controllers/agents/svc/serviceclass_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"time"
 
 	"go.uber.org/atomic"
@@ -30,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


controllers/agents/svc/serviceclass_controller.go
```diff
@@ -102,7 +103,14 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
-	if err = r.setOwnerReference(ctx, &serviceClass, req.Namespace); err != nil {
+	// get the controller's deployment
+	controller := appsv1.Deployment{}
+	controllerRef := types.NamespacedName{Namespace: serviceClass.Namespace, Name: constants.ServiceAgentDeploymentName}
+	if err = r.Get(ctx, controllerRef, &controller); err != nil {
+		reconcileLog.Error(err, "Failed to retrieve controller reference")
+	}
```
filariow, line 111: I'd say we should handle the error here

```suggestion
	if err := r.Get(ctx, controllerRef, &controller); err != nil {
		reconcileLog.Error(err, "Failed to retrieve controller reference")
		if apierrors.IsNotFound(err) {
			panic(err)
		}
		return ctrl.Result{}, err
	}
```

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -200,6 +176,46 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	return ctrl.Result{}, errors.Join(errs...)
 }
 
+func Update(ctx context.Context, remote_client client.Client, rs v1alpha1.RegisteredService, secret *v1.Secret) []error {
```
filariow, line 179: as this function is visible in the whole pkg, I'd propose to rename it to `updateRegisteredService`. The same suggestion applies for the above `Delete` function.

filariow (changes requested):
great approach! just some nits

controllers/agents/svc/serviceclass_controller.go
sadlerap: I'm not sure the panic is enough, since controller-runtime installs a panic handler around the reconciler.  In this case, should we call `os.Exit(1)`?

controllers/agents/svc/serviceclass_controller.go
sadlerap: Fixed.

controllers/agents/svc/serviceclass_controller.go
filariow: I'd go for `os.Exit(1)`. This way we are protecting the application from accidentally being installed with a different name. WDYT?

sadlerap: I'm not a huge fan of shutting down the entire pod; that seems like a great way to cause problems.  If there was a cleaner way of asking the controller to self-terminate, I'd be all for that.

filariow: I agree, we can find a better way for that. I'd propose to go with `os.Exit(1)` for now, add a `TODO` comment here, and open an issue for taking track.

controllers/agents/svc/serviceclass_controller.go
```diff
@@ -102,7 +104,21 @@ func (r *ServiceClassReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	controller := appsv1.Deployment{}
+	controllerRef := types.NamespacedName{Namespace: serviceClass.Namespace, Name: constants.ServiceAgentDeploymentName}
+	if err = r.Get(ctx, controllerRef, &controller); err != nil {
+		reconcileLog.Error(err, "Failed to retrieve controller reference")
+        if apierrors.IsNotFound(err) {
+            // FIXME(sadlerap): the deployment's been deleted, and the pod
+            // we're running in is likely going to be deleted soon as well.  Do
+            // we have a cleaner way of triggering our own shutdown?
+            os.Exit(1)
+        }
```
filariow, line 117: indentation fix and I'd propose to add a log line here

```suggestion
                if apierrors.IsNotFound(err) {
                    // FIXME(sadlerap): the deployment's been deleted, and the pod
                    // we're running in is likely going to be deleted soon as well.  Do
                    // we have a cleaner way of triggering our own shutdown?
                    reconcileLog.Error(err, "can not find agent's deployment", "agent", constants.ServiceAgentDeploymentName)
                    os.Exit(1)
                }
```

filariow: PTAL at #156

filariow (approved):
lgtm

PR #145, opened 2023-05-02
Handling catalog updates when service contraints change

dperaza4dustbit:
This fixes bug APPSVC-1339. 
Catalogs should change if constraints change and cause a previously matching Environment tag in a registered service to not match.

### DIFF
--- a/controllers/registeredservice_controller.go
+++ b/controllers/registeredservice_controller.go
@@ -99,7 +99,7 @@ func (r *RegisteredServiceReconciler) removeServiceFromCatalog(ctx context.Conte
 	return nil
 }
 
-func (r *RegisteredServiceReconciler) addServiceToCatalogs(ctx context.Context, rs primazaiov1alpha1.RegisteredService) error {
+func (r *RegisteredServiceReconciler) reconcileCatalogs(ctx context.Context, rs primazaiov1alpha1.RegisteredService) error {
 	log := log.FromContext(ctx)
 	catalogs, err := r.getServiceCatalogs(ctx, rs.Namespace)
 	if err != nil {
@@ -109,13 +109,22 @@ func (r *RegisteredServiceReconciler) addServiceToCatalogs(ctx context.Context, 
 
 	var errs []error
 	for _, sc := range catalogs.Items {
-		if rs.Spec.Constraints == nil || envtag.Match(sc.Name, rs.Spec.Constraints.Environments) {
+		if envtag.Match(sc.Name, rs.Spec.GetEnvironmentConstraints()) {
+			log.Info("Constraint matched or no constraints, reconciling catalog")
 			err = r.addServiceToCatalog(ctx, sc, rs)
 			if err != nil {
 				log.Error(err, "Error found adding RegisteredService to ServiceCatalog")
 				errs = append(errs, err)
 			}
 			log.Info("Added RegisteredService to ServiceCatalog", "RegisteredService", rs.Name, "ServiceCatalog", sc.Name)
+		} else {
+			log.Info("Constraint mismatched, reconciling catalog")
+			err = r.removeServiceFromCatalog(ctx, sc, rs.Namespace, rs.Name)
+			if err != nil {
+				log.Error(err, "Error found removing RegisteredService from ServiceCatalog")
+				errs = append(errs, err)
+			}
+
 		}
 	}
 
@@ -201,7 +210,7 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, err
 		}
 
-		err = r.addServiceToCatalogs(ctx, rs)
+		err = r.reconcileCatalogs(ctx, rs)
 
 		if err != nil {
 			// Service Catalog update failed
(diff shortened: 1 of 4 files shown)


test/acceptance/features/registerServiceConstraints.feature
```diff
@@ -49,6 +50,82 @@ Feature: Register a cloud service in Primaza cluster with constraints but withou
...
+        metadata:
+            name: worker
+            namespace: primaza-system
+        spec:
+            environmentName: dev
+            clusterContextSecret: primaza-kw
+        ---
+        apiVersion: primaza.io/v1alpha1
+        kind: RegisteredService
+        metadata:
```
Kartikey-star, line 71: l71-94 can be removed right

controllers/registeredservice_controller.go
```diff
@@ -109,13 +109,27 @@ func (r *RegisteredServiceReconciler) addServiceToCatalogs(ctx context.Context,
 
 	var errs []error
 	for _, sc := range catalogs.Items {
-		if rs.Spec.Constraints == nil || envtag.Match(sc.Name, rs.Spec.Constraints.Environments) {
+		constraints_match := true
```
Kartikey-star, line 112: Don't think L112-116 is required. We can simply use the earlier condition `if rs.Spec.Constraints == nil || envtag.Match(sc.Name, rs.Spec.Constraints.Environments)` , if true add service to catalog else remove service from catalog.

Kartikey-star (commented):
Looks good to me, just some minute changes to be considered

test/acceptance/features/registerServiceConstraints.feature
dperaza4dustbit: The idea here is to set it first to match so the catalog has the item, but when we do an update changing only the constraint to not match then the code added to remove from catalog is exercised. If we remove this then it will not exercise the removing part completely. Notice there are other test cases handling the not matching scenario on the create of the resource. This one focuses on the update, se L96

controllers/registeredservice_controller.go
dperaza4dustbit: Agree it will work as well, but I will prefer to be more explicit with the boolean to show what matches means. How strong do you feel about this one?

dperaza4dustbit (commented):
Thanks for review @Kartikey-star can you take a look at my comments?

test/acceptance/features/registerServiceConstraints.feature
Kartikey-star: Ok makes sense

Kartikey-star (dismissed):

controllers/registeredservice_controller.go
```diff
@@ -109,13 +109,27 @@ func (r *RegisteredServiceReconciler) addServiceToCatalogs(ctx context.Context,
...
+		if rs.Spec.Constraints != nil {
+			constraints_match = envtag.Match(sc.Name, rs.Spec.Constraints.Environments)
+		}
+
+		if constraints_match {
```
filariow, line 117: if you like it you can add for RegisteredService a method like [we did for ServiceClasses](https://github.com/primaza/primaza/blob/966880a5c65af36746d14d0bbec18a77da56f7b7/api/v1alpha1/serviceclass_types.go#LL89C1-L95C2) and refactor these lines as follows:

```suggestion
		if envtag.Match(sc.Name, rs.Spec.GetEnvironmentConstraints()) {
```

test/acceptance/features/registerServiceConstraints.feature
```diff
@@ -49,6 +50,82 @@ Feature: Register a cloud service in Primaza cluster with constraints but withou
...
+              value: mydavphost.io
+            - name: port
+              value: "5432"
+            - name: user
+              value: davp
+            - name: password
+              value: quedicelagente
+            - name: database
+              value: davpdata
+          sla: L3
```
filariow, line 94: we can move secret data out of RegisteredServices, applies also at L96

```suggestion
        ---
        apiVersion: v1
        kind: Secret
        metadata:
            name: $scenario_id
            namespace: primaza-system
        stringData:
            password: quedicelagente
        ---
        apiVersion: primaza.io/v1alpha1
        kind: RegisteredService
        metadata:
          name: primaza-rsdb
          namespace: primaza-system
        spec:
          constraints:
            environments:
              - dev
          serviceClassIdentity:
            - name: type
              value: psqlserver
            - name: provider
              value: aws
          serviceEndpointDefinition:
            - name: host
              value: mydavphost.io
            - name: port
              value: "5432"
            - name: user
              value: davp
            - name: password
              valueFromSecret:
                name: $scenario_id
                key: password
            - name: database
              value: davpdata
          sla: L3
```

filariow (changes requested):
some suggestions

controllers/registeredservice_controller.go
dperaza4dustbit: I like it, done

test/acceptance/features/registerServiceConstraints.feature
dperaza4dustbit: done

filariow (approved):
lgtm

PR #147, opened 2023-05-02
Refactor agent setup & teardown

sadlerap:
We can do a few things more efficiently:
- call Client.DeleteAllOf to delete all resources in a namespace without potentially having a TOCTOU bug
- filter deployments so that only the agent deployment triggers the reconciler.
- remove unused fields from the service agent struct


Kartikey-star (approved):
LGTM

filariow (approved):
nice! lgtm

PR #157, opened 2023-05-05
Control Plane: add runtime configuration

Kartikey-star:
Ref : https://issues.redhat.com/browse/APPSVC-1346


pkg/primaza/workercluster/agentapp.go
```diff
@@ -21,6 +21,7 @@ import (
 	_ "embed"
```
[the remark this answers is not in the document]
Kartikey-star: There seems to be one challenge to that `DeleteServiceAgent` and `DeleteApplicationAgent` method using the deployment names

cmd/primaza/main.go
```diff
@@ -47,6 +47,8 @@ const (
 	EnvHealthCheckInterval         = "HEALTH_CHECK_INTERVAL"
 	DefaultHealthCheckInterval int = 600
 	MinimumHealtCheckInterval  int = 10
+	EnvAppAgent                    = "AGENT_APP"
+	EnvSvcAgent                    = "AGENT_SVC"
```
filariow, line 51: I'd propose to add `Spec` in the name. Also applies in the other files

```suggestion
	EnvAppAgentSpec          = "AGENT_APP_SPEC"
	EnvSvcAgentSpec            = "AGENT_SVC_SPEC"
```

config/default/kustomization.yaml
filariow, line 1: this can be reverted

config/manager/configmap.yaml
```diff
@@ -6,3 +6,143 @@ data:
   agentsvc-image: agentsvc:latest
   agentapp-image: agentapp:latest
   health-check-interval: 10
+  # agentapp: application agent deployment will be created at runtime.
+  # agentsvc: service agent deployment will be created at runtime. Please refer to agents-configmap target for more information.
+
```
filariow, line 11: This can be removed

```suggestion
```

config/manager/kustomization.yaml
```diff
@@ -15,3 +15,7 @@ configMapGenerator:
   name: manager-config
   options:
     disableNameSuffixHash: true
+images:
+- name: controller
+  newName: primaza-controller
+  newTag: latest
```
filariow, line 21: this can be rolled-back

```suggestion
```

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/evanphx/json-patch/v5 v5.6.0 // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
 	github.com/go-logr/logr v1.2.4 // indirect
+	github.com/ghodss/yaml v1.0.0 // indirect
```
filariow, line 25: do we need it or can just use `gopkg.in/yaml`?

filariow (changes requested):
some nits

cmd/primaza/main.go
Kartikey-star: Now we are placing th entire deployment in the config map and not just the spec, so may be we shall use the name without `SPEC`

go.mod
Kartikey-star: This package is needed it seems , because with gopkg.in/yaml i get error on unmarshaling like `cannot unmarshal !!str `64Mi` into resource.Quantity`

Kartikey-star: Behind the scenes the package does use gopkg.in/yaml but converts the yaml to json and then json.unmarshal

cmd/primaza/main.go
filariow: good point. What about `Manifest`? I'd really like a more explicit variable name

go.mod
filariow: ok. I'd say we can use `https://github.com/kubernetes-sigs/yaml` as [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes/blob/15a1f9a39db95a7f83ce5463a328566411d99d51/go.mod#LL128C2-L128C18). It's a permanent fork of github.com/ghodss/yaml

filariow (approved):
lgtm

PR #158, opened 2023-05-05
automatically cancel superceded actions runs

sadlerap:
I've noticed that when we push changes to a branch, CI continues running against stale commits.  These runs are redundant, and they tie up our CI resources (which aren't infinite).

This automatically cancels runs with the same head ref (e.g. a PR), which should shorten our wait times slightly.  This does _not_ cancel runs for branch pushes (e.g. pushing to `main`).

Docs: https://docs.github.com/en/actions/using-jobs/using-concurrency


filariow (approved):
lgtm

PR #166, opened 2023-05-15
GitHub Pages documentation with mdbook

filariow:
For a result preview: https://filariow.github.io/primaza/


docs/book/book.toml
```diff
@@ -0,0 +1,6 @@
+[book]
+authors = ["Francesco Ilario"]
```
sadlerap, line 2: Maybe change this to `The Primaza Authors`

docs/book/src/architecture
sadlerap, line 1: I'm wondering if it makes sense to move the markdown files directly into this folder rather than use symlinks.  I don't expect to be consuming these files in places that we can't change the link location to (e.g. our readme), and I'd rather avoid symlinks in git repositories when possible (since I know some platforms have issues with them).

.github/workflows/pages.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Build book
+        run: cd docs/book && mdbook build
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v1
+        with:
+          # Upload the book
+          path: 'docs/book/book'
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v2
```
sadlerap, line 48: What URL will this deploy to?

.github/workflows/pages.yml
sadlerap, line 1: Should be `pages.yaml`, our linter isn't picking up this file.

sadlerap (commented):
Some minor feedback.  I really love how the pages look, but there are a few behind-the-scenes things we should change.

.github/workflows/pages.yml
filariow: it should be the repository's one: https://primaza.github.io/primaza

docs/book/src/architecture
filariow: agree, let me fix that

docs/book/book.toml
filariow: great catch!

.github/workflows/pages.yml
filariow: great catch, fixed!

PR #176, opened 2023-05-20
Add Synchronization Strategy

filariow:
Depends on "Fix Cluster Environment healthchecks" #175

Enables pull strategy for Primaza.
Pull Strategy can be enabled for each Cluster Environments.
Control Plane watches RegisteredServices in Service Namespaces and ServiceClaims in Application Namespaces.
Pull Strategy won't need the Agents to being able to push resources to Primaza's Control Plane.
This allows higher security, but lower performances.


Kartikey-star (approved):
Looks good to me.

PR #193, opened 2023-05-23
don't run acceptance tests on push to main

sadlerap:
Since we enforce that PRs are up-to-date before they can be merged, this is a redundant check.


filariow (approved):
Lgtm

PR #201, opened 2023-06-07
Update README

filariow:


README.md
```diff
@@ -1,65 +1,24 @@
-# Primaza
+# :knot: Primaza
 
-## Multi Cluster Service Consumption Framework
+Primaza is a multi-cluster Service Consumption Framework.
+Primaza wants to be namespace-scoped and will not required any resource at cluster level.
```
dperaza4dustbit, line 4: suggest to change this one: Primaza is namespace-scoped and does not require any resource at cluster level.

dperaza4dustbit (commented):
suggested a change but look good

sadlerap (approved):

README.md
sadlerap: We definitely require CRDs, which are cluster scoped, and our webhooks are (currently) cluster-scoped as well.  Maybe we should mention this?

filariow: Added mention to CRDs. We also already have a [story for removing webhooks](https://issues.redhat.com/browse/APPSVC-1363), so I'd say we can omit them here.

dperaza4dustbit: look good to me

dperaza4dustbit (approved):
lgtm

PR #208, opened 2023-06-13
Add Release workflow

filariow:
Please take a look at https://github.com/filariow/primaza/actions/runs/5259332688 and https://github.com/filariow/primaza/releases/tag/v0.1.0

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,140 @@
+name: Release primaza
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    tags:
+    - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Primaza release version"
+        required: true
+        default: "latest"
+
+permissions:
+  contents: write
+
+env:
+  GO111MODULE: "on"
+  GO_VERSION: "^1.20"
+  OUT_DIR: ./out/workflow
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build:
+    name: Release Primaza
+    runs-on: ubuntu-20.04
+    permissions:
+      contents: write
+      packages: write
+
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Set Version and Manifests
+        run: |
+          # set version based on dispatch type
+          version="latest"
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,140 @@
+name: Release primaza
+
+on:
```
sadlerap, line 3:
```suggestion
on: # yamllint disable-line rule:truthy
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+      - name: Delete previous release and tag
+        if: ${{ env.VERSION == 'latest' }}
+        uses: dev-drprasad/delete-tag-and-release@v0.2.1
+        with:
+          delete_release: true # default: false
+          tag_name: ${{ env.VERSION }} # tag name to delete
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
```
sadlerap, line 74: I don't know if deleting the tag is a good idea.  Once created, tags are going to get pulled by anyone with a repository clone, and once tags have been pulled by those clones, they can't really be deleted.

For more information, see [git's documentation on re-tagging](https://git-scm.com/docs/git-tag#_on_re_tagging).

.github/workflows/release.yaml
```diff
@@ -0,0 +1,140 @@
...
+          if [[ "${{github.event_name}}" == "workflow_dispatch" ]]; then
+            version=${{ github.event.inputs.version }}
+          elif [[ "${{github.ref_type}}" == "tag" ]]; then
+            version=${{ github.ref_name }}
+          fi
+          echo "VERSION=$version" >> $GITHUB_ENV
+          echo "CONTROL_PLANE_CONFIG_FILE=control_plane_config_$version.yaml" >> $GITHUB_ENV
+          echo "CRDS_CONFIG_FILE=crds_config_$version.yaml" >> $GITHUB_ENV
+          echo "APPLICATION_NAMESPACE_CONFIG_FILE=application_namespace_config_$version.yaml" >> $GITHUB_ENV
+          echo "SERVICE_NAMESPACE_CONFIG_FILE=service_namespace_config_$version.yaml" >> $GITHUB_ENV
```
mmulholla, line 51: are these config files meant for consumption by primazactl? At the moment primazactl release process creates its own versions which it uses. If it should use these some changes to primazactl will be needed.

mmulholla (commented):
LGTM but have a question.

.github/workflows/release.yaml
filariow: I'd say primazactl should use manifests from primaza's repo so that manifests are versioned together with Primaza releases

mmulholla (approved):

PR #210, opened 2023-06-14
Reconcile on ClusterContext Secret changes

filariow:
When the ClusterEnvironment's ClusterContext secret is updated, its ClusterEnvironment is reconciled

Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #214, opened 2023-06-15
Fix healthcheck crash on invalid kubeconfig

baijum:
Ref. https://issues.redhat.com/browse/APPSVC-1372

### DIFF
--- a/controllers/clusterenvironment_controller.go
+++ b/controllers/clusterenvironment_controller.go
@@ -341,7 +341,7 @@ func (r *ClusterEnvironmentReconciler) MonitorHealth(ctx context.Context, ns str
 			ce := &(ceList.Items[i])
 
 			// get cluster config
-			cfg, err := clustercontext.GetClusterRESTConfig(ctx, r.Client, ce.Namespace, ce.Spec.ClusterContextSecret)
+			cfg, err := r.retrieveClusterContextSecret(ctx, ce)
 			if err != nil {
 				if errors.Is(err, clustercontext.ErrSecretNotFound) {
 					c := workercluster.ConnectionStatus{
@@ -354,6 +354,19 @@ func (r *ClusterEnvironmentReconciler) MonitorHealth(ctx context.Context, ns str
 						l.Error(err, "error updating cluster environment status", "status", ce.Status)
 					}
 				}
+			}
+			if cfg == nil {
+				l.Info("Not running healthchecks for ClusterEnvironment: empty kubeconfig from ClusterContext Secret", "ClusterEnvironment", ce.Name)
+				c := workercluster.ConnectionStatus{
+					State:   primazaiov1alpha1.ClusterEnvironmentStateOffline,
+					Reason:  ErrorDuringHealthCheckReason,
+					Message: "kubeclient config is empty",
+				}
+				r.updateClusterEnvironmentStatus(ctx, ce, c)
+				if err := r.Client.Status().Update(ctx, ce); err != nil {
+					l.Error(err, "error updating cluster environment status", "status", ce.Status)
+				}
+				continue
 			}
 
 			// test connection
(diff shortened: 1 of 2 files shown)


Kartikey-star (approved):
lgtm, Checking whether the config being returned is empty can  be an additional check  after https://github.com/primaza/primaza/pull/210/ is merged or either this PR is merged. This condition and status updates can be added in `extractClusterContextRESTConfig` method. WDYT?

controllers/clusterenvironment_controller.go
```diff
@@ -278,6 +278,20 @@ func (r *ClusterEnvironmentReconciler) MonitorHealth(ctx context.Context, ns str
 					}
 				}
 			}
+			if cfg == nil {
+				l.Error(errors.New("Empty conig"), "kubeclient config is empty")
```
filariow, line 282:
```suggestion
				l.Info("Not running healthchecks for ClusterEnvironment: empty kubeconfig from ClusterContext Secret", "ClusterEnvironment", ce.Name)
```

controllers/clusterenvironment_controller.go
```diff
@@ -278,6 +278,20 @@ func (r *ClusterEnvironmentReconciler) MonitorHealth(ctx context.Context, ns str
...
+				c := workercluster.ConnectionStatus{
+					State:   primazaiov1alpha1.ClusterEnvironmentStateOffline,
+					Reason:  ErrorDuringHealthCheckReason,
+					Message: "kubeclient config is empty",
+				}
+				r.updateClusterEnvironmentStatus(ctx, ce, c)
+				if err := r.Client.Status().Update(ctx, ce); err != nil {
+					l.Error(err, "error updating cluster environment status", "status", ce.Status)
+				}
+				time.Sleep(time.Duration(hci) * time.Second)
```
filariow, line 292: Is this needed? I'd say we can continue directly with next ClusterEnvironment in the list and sleep at L333

```suggestion
```

controllers/clusterenvironment_controller.go
```diff
@@ -278,6 +278,20 @@ func (r *ClusterEnvironmentReconciler) MonitorHealth(ctx context.Context, ns str
 					}
 				}
 			}
+			if cfg == nil {
```
filariow, line 281: we may handle this check in `clustercontext.GetClusterRESTConfig` and return an error from there. WDYT?

filariow (changes requested):
some ideas

controllers/clusterenvironment_controller.go
baijum: There are other places where this function is called where empty config is compatible. So, I would prefer to retain the current code. I will work on the other suggestions.

controllers/clusterenvironment_controller.go
baijum: Done.

controllers/clusterenvironment_controller.go
baijum: Done.

controllers/clusterenvironment_controller.go
filariow: Right, good point!

baijum:
> lgtm, Checking whether the config being returned is empty can be an additional check after #210 is merged or either this PR is merged. This condition and status updates can be added in `extractClusterContextRESTConfig` method. WDYT?

@Kartikey-star 

Now PR #210 has been merged. I have updated this PR to call  `extractClusterContextRESTConfig` method. Since empty config is compatible in other places where that function is called, I have kept the check outside the function.

controllers/clusterenvironment_controller.go
```diff
@@ -735,7 +748,7 @@ func (r *ClusterEnvironmentReconciler) CreateServiceCatalog(ctx context.Context,
 }
 
 func (r *ClusterEnvironmentReconciler) finalizeClusterEnvironmentInNamespaces(ctx context.Context, ce *primazaiov1alpha1.ClusterEnvironment) error {
-	kcfg, err := clustercontext.GetClusterRESTConfig(ctx, r.Client, ce.Namespace, ce.Spec.ClusterContextSecret)
+	kcfg, err := r.retrieveClusterContextSecret(ctx, ce)
```
baijum, line 751: This is an unrelated change. I made this change based on PR #210
Let me know if I should send a separate PR.

filariow (approved):
lgtm

PR #216, opened 2023-06-15
Updating registered service triggers service claim reconciliation

Kartikey-star:
Refer: https://issues.redhat.com/browse/APPSVC-1366

### DIFF
--- a/controllers/serviceclaim_controller.go
+++ b/controllers/serviceclaim_controller.go
@@ -30,7 +30,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/google/uuid"
 	"github.com/primaza/primaza/api/v1alpha1"
@@ -115,17 +118,14 @@ func (r *ServiceClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	case "":
 		sclaim.Status.ClaimID = uuid.New().String()
 		l.Info("reconciling new service claim")
-		return ctrl.Result{}, r.processPendingClaim(ctx, req, sclaim)
-	case primazaiov1alpha1.ServiceClaimStatePending:
-		l.Info("reconciling pending service claim")
-		return ctrl.Result{}, r.processPendingClaim(ctx, req, sclaim)
+		return ctrl.Result{}, r.processClaim(ctx, req, sclaim)
 	default:
-		l.Info("reconciling resolved service claim")
-		return ctrl.Result{}, nil
+		l.Info("reconciling resolved and pending service claim")
+		return ctrl.Result{}, r.processClaim(ctx, req, sclaim)
 	}
 }
 
-func (r *ServiceClaimReconciler) processPendingClaim(ctx context.Context, req ctrl.Request, sclaim primazaiov1alpha1.ServiceClaim) error {
+func (r *ServiceClaimReconciler) processClaim(ctx context.Context, req ctrl.Request, sclaim primazaiov1alpha1.ServiceClaim) error {
 	l := log.FromContext(ctx)
 
 	var rsl primazaiov1alpha1.RegisteredServiceList
@@ -488,7 +488,40 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ServiceClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	genPred := predicate.GenerationChangedPredicate{}
+	reconcileOnRegisteredServiceUpdate := func(ctx context.Context, a client.Object) []reconcile.Request {
+		l := log.FromContext(ctx)
+		rs, ok := a.(*v1alpha1.RegisteredService)
+		if !ok {
+			l.Info("error parsing object to RegisteredService when mapping to ServiceClaim reconciliation trigger", "object", a)
+			return []reconcile.Request{}
+		}
+		if rs.Status.State != `Claimed` {
+			l.Info("Registered service is unclaimed, no service claim to reconcile", "registered-service", rs.Name)
+			return []reconcile.Request{}
+		}
+		serviceclaims := &v1alpha1.ServiceClaimList{}
+		opts := &client.ListOptions{}
+		if err := r.List(ctx, serviceclaims, opts); err != nil {
+			l.Error(err,
+				"unable to list the ServiceClaims and reconcile for Registered Service Updates",
+				"RegisteredService", rs.Name)
+			return []reconcile.Request{}
+		}
+		for _, sc := range serviceclaims.Items {
+			if sc.Status.State == "Resolved" && sc.Status.RegisteredService == rs.Name {
+				return []reconcile.Request{{NamespacedName: types.NamespacedName{
+					Namespace: sc.Namespace,
+					Name:      sc.Name,
+				}}}
+			}
+		}
+		return []reconcile.Request{}
+	}
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&primazaiov1alpha1.ServiceClaim{}).
+		Watches(&primazaiov1alpha1.RegisteredService{}, handler.EnqueueRequestsFromMapFunc(reconcileOnRegisteredServiceUpdate)).
+		WithEventFilter(genPred).
 		Complete(r)
 }
--- a/test/acceptance/features/serviceClaim.feature
+++ b/test/acceptance/features/serviceClaim.feature
@@ -158,7 +158,7 @@ Feature: Service claim with label selector
         And  On Primaza Cluster "main", RegisteredService "primaza-rsdb" state will eventually move to "Claimed"
         And  On Primaza Cluster "main", ServiceCatalog "stage" will not contain RegisteredService "primaza-rsdb"
         And  On Worker Cluster "worker", the secret "sc-test" in namespace "applications" has the key "type" with value "psqlserver"
-  
+
     Scenario: Create a service claim with label selector, no constraints, sci subset and sed subset
         Given On Primaza Cluster "main", Resource is created
         """
@@ -216,7 +216,7 @@ Feature: Service claim with label selector
         And On Primaza Cluster "main", RegisteredService "primaza-rsdb" state will eventually move to "Claimed"
         And On Primaza Cluster "main", ServiceCatalog "stage" will not contain RegisteredService "primaza-rsdb"
         And  On Worker Cluster "worker", the secret "sc-test" in namespace "applications" has the key "type" with value "psqlserver"
-  
+
     Scenario: Create a service claim with non-existing SED key
         Given On Primaza Cluster "main", Resource is created
         """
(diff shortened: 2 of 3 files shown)


controllers/serviceclaim_controller.go
```diff
@@ -486,9 +486,47 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+
+	genPred := predicate.GenerationChangedPredicate{}
+	reconcileOnRegisteredServiceUpdate := func(ctx context.Context, a client.Object) []reconcile.Request {
+		l := log.FromContext(ctx)
+		serviceclaims := &v1alpha1.ServiceClaimList{}
+		opts := &client.ListOptions{}
+		if err := r.List(ctx, serviceclaims, opts); err != nil {
+			return []reconcile.Request{}
+		}
+		rs := a.(*v1alpha1.RegisteredService)
```
filariow, line 500: this should not happen as of now, but it may be better to check whether the casting is successful:

```suggestion
		rs, ok := a.(*v1alpha1.RegisteredService)
		if !ok {
			l.Info("error parsing object to RegisteredService when mapping to ServiceClaim reconciliation trigger", "object", a)
			return []reconcile.Request{}
		}
```

controllers/serviceclaim_controller.go
```diff
@@ -486,9 +486,47 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
 	return errors.Join(errs...)
 }
 
-// SetupWithManager sets up the controller with the Manager.
+// SetupWithManager setup controller with manager
```
filariow, line 489: this can be discarded

controllers/serviceclaim_controller.go
```diff
@@ -486,9 +486,47 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+		for _, sc := range serviceclaims.Items {
+			env := sc.Spec.EnvironmentTag
+			if sc.Spec.ApplicationClusterContext != nil {
+				var err error
+				ce, err := r.getEnvironmentFromClusterEnvironment(ctx, reconcile.Request{NamespacedName: types.NamespacedName{
+					Namespace: sc.Namespace,
+					Name:      sc.Name,
+				}}, sc.Spec.ApplicationClusterContext.ClusterEnvironmentName)
+				if err != nil {
+					l.Error(err, "unable to get environment from cluster environment")
```
filariow, line 511:
```suggestion
					l.Error(err, "unable to get environment from cluster environment",
						"cluster-environment", ce.Name,
						"environment", sc.Spec.ApplicationClusterContext.ClusterEnvironmentName)
						
```

controllers/serviceclaim_controller.go
```diff
@@ -486,9 +486,47 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+		}
+		rs := a.(*v1alpha1.RegisteredService)
+		reply := make([]reconcile.Request, 0, len(serviceclaims.Items))
+		for _, sc := range serviceclaims.Items {
+			env := sc.Spec.EnvironmentTag
+			if sc.Spec.ApplicationClusterContext != nil {
+				var err error
+				ce, err := r.getEnvironmentFromClusterEnvironment(ctx, reconcile.Request{NamespacedName: types.NamespacedName{
+					Namespace: sc.Namespace,
+					Name:      sc.Name,
```
filariow, line 508: Name is not used by the `getEnvironmentFromClusterEnvironment` func

```suggestion
```

test/acceptance/features/serviceClaim.feature
filariow, line 1: should we add also a scenario for value updated in a secret referenced by the RegisteredService? Are we planning this in a new PR?

filariow (changes requested):
some nits

test/acceptance/features/serviceClaim.feature
filariow: You're right. I'd propose to go with two stories

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +488,50 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+		for _, sc := range serviceclaims.Items {
+			env := sc.Spec.EnvironmentTag
+			if sc.Spec.ApplicationClusterContext != nil {
+				var err error
+				ce, err := r.getEnvironmentFromClusterEnvironment(ctx, reconcile.Request{NamespacedName: types.NamespacedName{
+					Namespace: sc.Namespace,
+				}}, sc.Spec.ApplicationClusterContext.ClusterEnvironmentName)
+				if err != nil {
+					l.Error(err, "unable to get environment from cluster environment",
+						"cluster-environment", ce.Name,
```
filariow, line 515: I'm expecting `ce` to be `nil` here, this log should panic

```suggestion
```

controllers/serviceclaim_controller.go
```diff
@@ -116,12 +119,9 @@ func (r *ServiceClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 		l.Info("reconciling new service claim")
 		return ctrl.Result{}, r.processPendingClaim(ctx, req, sclaim)
-	case primazaiov1alpha1.ServiceClaimStatePending:
-		l.Info("reconciling pending service claim")
-		return ctrl.Result{}, r.processPendingClaim(ctx, req, sclaim)
 	default:
-		l.Info("reconciling resolved service claim")
-		return ctrl.Result{}, nil
+		l.Info("reconciling resolved and pending service claim")
+		return ctrl.Result{}, r.processPendingClaim(ctx, req, sclaim)
```
filariow, line 124: there is a mismatch between log line and function name. Are we processing resolved and pending or just pending claims?

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +488,50 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+			return []reconcile.Request{}
+		}
+		rs, ok := a.(*v1alpha1.RegisteredService)
+		if !ok {
+			l.Info("error parsing object to RegisteredService when mapping to ServiceClaim reconciliation trigger", "object", a)
+			return []reconcile.Request{}
+		}
+		reply := make([]reconcile.Request, 0, len(serviceclaims.Items))
+		for _, sc := range serviceclaims.Items {
+			env := sc.Spec.EnvironmentTag
```
filariow, line 507: can we just use the `status.registeredService` field to decide if we need to reconcile the ServiceClaim?

filariow (commented):
some comments

controllers/serviceclaim_controller.go
Kartikey-star: We are resolving both resolved as well as pending service claims here. For a resolved claim if the registered service is updated, i am using this method to extract the sed. Makes sense to rename it though, just `processClaim`

controllers/serviceclaim_controller.go
Kartikey-star: Yeah by doing taht we can avoid the sci matching being duplicated

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +488,34 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
 // SetupWithManager sets up the controller with the Manager.
 func (r *ServiceClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	genPred := predicate.GenerationChangedPredicate{}
+	reconcileOnRegisteredServiceUpdate := func(ctx context.Context, a client.Object) []reconcile.Request {
+		l := log.FromContext(ctx)
+		serviceclaims := &v1alpha1.ServiceClaimList{}
+		opts := &client.ListOptions{}
+		if err := r.List(ctx, serviceclaims, opts); err != nil {
+			return []reconcile.Request{}
```
filariow, line 498: we should add a log here to tell the administrator that the registered service related service claim could not be reconciled

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +488,34 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+		}
+		rs, ok := a.(*v1alpha1.RegisteredService)
+		if !ok {
+			l.Info("error parsing object to RegisteredService when mapping to ServiceClaim reconciliation trigger", "object", a)
+			return []reconcile.Request{}
+		}
+		reply := make([]reconcile.Request, 0, len(serviceclaims.Items))
+		if rs.Status.State == `Claimed` {
```
filariow, line 506: this check can be moved on top of the method. Before doing any other action we can check if the registered service is claimed, if not we don't need to reconcile anything

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +488,34 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+			for _, sc := range serviceclaims.Items {
+				reply = append(reply, reconcile.Request{NamespacedName: types.NamespacedName{
+					Namespace: sc.Namespace,
+					Name:      sc.Name,
+				}})
+			}
```
filariow, line 512: that's not clear to me. I was expecting something like this

```suggestion
			for _, sc := range serviceclaims.Items {
				if sc.Status != nil && sc.Status.State == "Resolved" && sc.Status.RegisteredService == rs.Name {
					return []reconcile.Request{
						{
							NamespacedName: types.NamespacedName{
								Namespace: sc.Namespace,
								Name:      sc.Name,
							},
						},
					}
				}
			}
			return []reconcile.Request{}
```

filariow (changes requested):
some nits

controllers/serviceclaim_controller.go
Kartikey-star: Makes sense

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +489,41 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+		}
+		serviceclaims := &v1alpha1.ServiceClaimList{}
+		opts := &client.ListOptions{}
+		if err := r.List(ctx, serviceclaims, opts); err != nil {
+			l.Error(err,
+				"unable to list the ServiceClaims and reconcile for Registered Service Updates",
+				"RegisteredService", rs.Name)
+			return []reconcile.Request{}
+		}
+		reply := make([]reconcile.Request, 0, len(serviceclaims.Items))
```
filariow, line 513: ServiceClaim - RegisteredService relationship is one to one. I don't think we need to look for more then one Service Claim. right?

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +489,41 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+		for _, sc := range serviceclaims.Items {
+			if !reflect.DeepEqual(sc.Status, v1alpha1.ServiceClaimStatus{}) && sc.Status.State == "Resolved" && sc.Status.RegisteredService == rs.Name {
```
filariow, line 515: I think we can simplify this check. WDYT?

```suggestion
			if sc.Status.State == "Resolved" && sc.Status.RegisteredService == rs.Name {
```

Kartikey-star: Can the sc.Status be empty?.. It would result in panic in that case.

filariow: we have panic only if sc.Status is nil, but it is not possible as it is not a pointer

```go
type ServiceClaim struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   ServiceClaimSpec   `json:"spec,omitempty"`
	Status ServiceClaimStatus `json:"status,omitempty"`
}
```

controllers/serviceclaim_controller.go
```diff
@@ -488,7 +488,42 @@ func (r *ServiceClaimReconciler) DeleteServiceBindingsAndSecret(
...
+			return []reconcile.Request{}
+		}
+		reply := make([]reconcile.Request, 0, 1)
+		for _, sc := range serviceclaims.Items {
+			if sc.Status.State == "Resolved" && sc.Status.RegisteredService == rs.Name {
+				reply = append(reply, reconcile.Request{NamespacedName: types.NamespacedName{
+					Namespace: sc.Namespace,
+					Name:      sc.Name,
+				}})
+				return reply
```
filariow, line 519: this can be simplified, other than this LGTM
```suggestion
		for _, sc := range serviceclaims.Items {
			if sc.Status.State == "Resolved" && sc.Status.RegisteredService == rs.Name {
				return []reconcile.Request{{NamespacedName: types.NamespacedName{
					Namespace: sc.Namespace,
					Name:      sc.Name,
				}}})
```

filariow (changes requested):

filariow (approved):
LGTM

PR #218, opened 2023-06-19
Check for excess permission in ClusterEnvironment

baijum:
Ref. https://issues.redhat.com/browse/APPSVC-1309


hack/permissionlist/permissionlist.go
```diff
@@ -0,0 +1,71 @@
+package main
+
+import (
+	"fmt"
+	"html/template"
```
baijum, line 5: change it to `text/template`

pkg/authz/permission_list_app.go
```diff
@@ -0,0 +1,76 @@
...
+		Verbs:         []string{"get", "list", "watch", "update", "patch", "delete", "deletecollection"},
+	})
+	AppPermissionList = append(AppPermissionList, Permission{
+		APIGroups:     []string{"primaza.io"},
+		Resources:     []string{"serviceclaims/status"},
+		ResourceNames: []string{},
+		Namespace:     "system",
+		Name:          "primaza:app:manager",
+		Verbs:         []string{"get", "list", "watch", "update"},
+	})
```
filariow, line 19:
```suggestion
var AppPermissionList = []Permission{
	{
		APIGroups:     []string{"primaza.io"},
		Resources:     []string{"servicebindings", "serviceclaims", "servicecatalogs"},
		ResourceNames: []string{},
		Namespace:     "system",
		Name:          "primaza:app:manager",
		Verbs:         []string{"get", "list", "watch", "update", "patch", "delete", "deletecollection"},
	},
	{
		APIGroups:     []string{"primaza.io"},
		Resources:     []string{"serviceclaims/status"},
		ResourceNames: []string{},
		Namespace:     "system",
		Name:          "primaza:app:manager",
		Verbs:         []string{"get", "list", "watch", "update"},
	},
```

hack/permissionlist/permissionlist.go
filariow, line 1: may you add some comments to this file?

cmd/primaza/main.go
filariow, line 1: We should also check that generated files are up to date in the PR Check pipeline.
I'd propose to add `make generate` at https://github.com/primaza/primaza/blob/f2959bc4a65cf32bda5157d20b71ed8f9bfa3606/.github/workflows/pr-checks.yaml#L82C21-L82C21

controllers/clusterenvironment_controller.go
```diff
@@ -172,12 +172,20 @@ func (r *ClusterEnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	// test connection
 	if err := r.testConnection(ctx, cfg, ce); err != nil {
+		l.Error(err, "error testing connection")
 		return ctrl.Result{}, err
 	}
 
+	// check excess permissions
+	err = r.checkExcessPermissions(ctx, cfg, ce)
+	if err != nil {
```
filariow, line 181:
```suggestion
	if err := r.checkExcessPermissions(ctx, cfg, ce); err != nil {
```

pkg/authz/permissions.go
```diff
@@ -79,3 +93,93 @@ func checkAccess(ctx context.Context, c *kubernetes.Clientset, np NamespacedPerm
...
+		if len(APIGroups) == 0 {
+			APIGroups = []string{""}
+		}
+		ResourceNames := rr.ResourceNames
+		if len(ResourceNames) == 0 {
+			ResourceNames = []string{""}
+		}
+		for _, ag := range APIGroups {
+			for _, rs := range rr.Resources {
+				for _, rn := range ResourceNames {
```
filariow, line 124: If `ResourceNames` is empty we are not executing the next loops right?

pkg/authz/permissions.go
```diff
@@ -79,3 +93,93 @@ func checkAccess(ctx context.Context, c *kubernetes.Clientset, np NamespacedPerm
...
+		ObjectMeta: metav1.ObjectMeta{Namespace: ns},
+		Spec:       authzv1.SelfSubjectRulesReviewSpec{Namespace: ns},
+	}
+
+	r, err := c.AuthorizationV1().SelfSubjectRulesReviews().Create(ctx, &ssrr, metav1.CreateOptions{})
+	if err != nil {
+		return []string{}, err
+	}
+	resourceRules := r.Status.ResourceRules
+	for _, rr := range resourceRules {
```
filariow, line 113: can we reduce the number of nested `for` loops?

filariow (changes requested):
some nits

make/primaza/build.mk
```diff
@@ -4,6 +4,9 @@ PRIMAZA_MAIN=./cmd/primaza/main.go
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
+	$(GO) generate
+	$(GO) fmt pkg/authz/permission_list_app.go
+	$(GO) fmt pkg/authz/permission_list_svc.go
```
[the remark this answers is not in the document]
baijum: Done

pkg/authz/permissions.go
baijum: See the line no. 119 to 121. I am adding an empty string to the slice.

pkg/authz/permission_list_app.go
baijum: I prefer to retain the current approach of initializing `AppPermissionList`.  Few reasons:

1. There is a clear separation between the variable declaration (`pkg/authz/permissions.go`) and initialization.
2. Since `init` is called once during runtime, there is no performance issue.
3. The `pkg/authz/permission_list_app.go` is a generated code. In fact, we need not commit this script 
   in the source tree. The program builds even without the generated files.
4. I don't see any advantage in declaring and initializing the `AppPermissionList` in the generated code. In fact,
   the program won't build without the generated code, if we take this approach.

hack/permissionlist/permissionlist.go
baijum: Done.

pkg/authz/permissions.go
```diff
@@ -79,3 +93,93 @@ func checkAccess(ctx context.Context, c *kubernetes.Clientset, np NamespacedPerm
 
 	return r.Status.Allowed, nil
 }
+
+func creatUnwantedPermissionsList(ctx context.Context, cfg *rest.Config, unwantedPermissions []string, ns string, definedRules map[[4]string]int) ([]string, error) {
```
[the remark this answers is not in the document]
baijum: Created a new function and added test for the same.

pkg/authz/permissions.go
baijum: I have created a new function with unit tests.

cmd/primaza/main.go
baijum: Done

pkg/authz/permission_list_app.go
filariow: 1. This is a PRO for sure, but to me this not justifies the CONS that follows
2. `init` is called when the module is loaded. This side-effect on module import may be really annoying and hard to test, reduce readability and increase complexity (cfr. https://github.com/leighmcculloch/gochecknoinits)
3. I'd suggest to commit the generated code. This will give the repo reader a full understanding of what's really happening.
4. I'm expecting the code to not build without generating code and checks on pipeline to be sure to have the latest generated code. This way what you see in the repo is matching exactly what it's built for a given version of the binary

baijum: Thanks for the detailed explanation. I agree with your points. Let me make the necessary changes.

baijum: Done

pkg/authz/permissions.go
```diff
@@ -79,3 +90,100 @@ func checkAccess(ctx context.Context, c *kubernetes.Clientset, np NamespacedPerm
...
+			}
+			for _, ag := range APIGroups {
+				for _, rs := range p.Resources {
+					for _, rn := range ResourceNames {
+						for _, vb := range p.Verbs {
+							definedRules[[4]string{ag, rs, rn, vb}] = 1
+						}
+					}
+				}
+			}
```
filariow, line 178: can we just refactor as follows:

```suggestion
			if len(p.APIGroups) == 0 || len(p.ResourceNames) == 0 {
				continue
			}
			for _, ag := range p.APIGroups {
				for _, rs := range p.Resources {
					for _, rn := range p.ResourceNames {
						for _, vb := range p.Verbs {
							definedRules[[4]string{ag, rs, rn, vb}] = 1
						}
					}
				}
			}
```

cmd/primaza/main.go
```diff
@@ -14,6 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//go:generate go run hack/permissionlist/permissionlist.go App pkg/authz/ config/agents/app/rbac/manager_role.yaml config/agents/app/rbac/leader_election_role.yaml
+//go:generate go run hack/permissionlist/permissionlist.go Svc pkg/authz/ config/agents/svc/rbac/manager_role.yaml config/agents/svc/rbac/leader_election_role.yaml
+//go:generate go fmt pkg/authz/permission_list_app.go
+//go:generate go fmt pkg/authz/permission_list_svc.go
```
filariow, line 20: I'd propose to have have them on the same line

```suggestion
//go:generate go run hack/permissionlist/permissionlist.go App pkg/authz/ config/agents/app/rbac/manager_role.yaml config/agents/app/rbac/leader_election_role.yaml && go fmt pkg/authz/permission_list_app.go
//go:generate go run hack/permissionlist/permissionlist.go Svc pkg/authz/ config/agents/svc/rbac/manager_role.yaml config/agents/svc/rbac/leader_election_role.yaml && go fmt pkg/authz/permission_list_svc.go
```

controllers/clusterenvironment_controller.go
```diff
@@ -534,6 +546,52 @@ func (r *ClusterEnvironmentReconciler) reconcileApplicationNamespaces(ctx contex
...
+			errs = append(errs, err)
+		}
+	}
+
+	// check service namespaces permissions
+	spc := controlplane.NewAgentSvcPermissionsChecker(cfg)
+	ep, err = spc.CheckExcessPermission(ctx, ce.Spec.ServiceNamespaces)
+	if err != nil {
+		errs = append(errs, err)
+	} else if len(ep) > 0 {
```
filariow, line 578: Just a refactor proposal

```suggestion
	errs := []error{}

	// check application namespaces permissions
	apc := controlplane.NewAgentAppPermissionsChecker(cfg)
	if ep, err := apc.CheckExcessPermission(ctx, ce.Spec.ApplicationNamespaces); err != nil {
		errs = append(errs, err)
	} else if len(ep) > 0 {
		m := metav1.Condition{
			Type:    "ExcessPermissions",
			Reason:  "ExcessPermissions",
			Message: "More permissions than required granted for Application Namespaces",
			Status:  "True",
		}

		meta.SetStatusCondition(&ce.Status.Conditions, m)
		if err := r.Client.Status().Update(ctx, ce); err != nil {
			l.Error(err, "error updating cluster environment status", "status", ce.Status)
			errs = append(errs, err)
		}
	}

	// check service namespaces permissions
	spc := controlplane.NewAgentSvcPermissionsChecker(cfg)
	if ep, err = spc.CheckExcessPermission(ctx, ce.Spec.ServiceNamespaces); err != nil {
		errs = append(errs, err)
	} else if len(ep) > 0 {
```

filariow (changes requested):
mostly good to me, just some nits

pkg/authz/permissions.go
baijum: We must iterate once when the length is zero with an empty string as a value. Continuing the loop will skip some of the rules.

filariow: ok

filariow (approved):
lgtm

PR #223, opened 2023-06-21
Service Binding Environment Variable Support

Kartikey-star:
Reference: https://issues.redhat.com/browse/APPSVC-1371

### DIFF
--- a/controllers/agents/app/servicebinding_controller.go
+++ b/controllers/agents/app/servicebinding_controller.go
@@ -419,6 +419,17 @@ func (r *ServiceBindingReconciler) getApplication(ctx context.Context,
 	return applications, nil
 }
 
+func removeServiceBindingEnvironments(envList []v1.EnvVar, sb primazaiov1alpha1.ServiceBinding) []v1.EnvVar {
+	var envListCopy []v1.EnvVar
+	for _, val := range envList {
+		if val.ValueFrom != nil && val.ValueFrom.SecretKeyRef != nil &&
+			val.ValueFrom.SecretKeyRef.Name != sb.Spec.ServiceEndpointDefinitionSecret {
+			envListCopy = append(envListCopy, val)
+		}
+	}
+	return envListCopy
+}
+
 func (r *ServiceBindingReconciler) updateContainerInfo(ctx context.Context, containers []interface{}, sb primazaiov1alpha1.ServiceBinding, mountPathDir, volumeName string, psSecret *v1.Secret) error {
 
 	l := log.FromContext(ctx)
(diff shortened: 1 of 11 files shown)


controllers/agents/app/servicebinding_controller.go
```diff
@@ -432,12 +441,30 @@ func (r *ServiceBindingReconciler) updateContainerInfo(ctx context.Context, cont
...
+					Name: e.Name,
+					ValueFrom: &v1.EnvVarSource{
+						SecretKeyRef: &v1.SecretKeySelector{
+							Key: e.Key,
+							LocalObjectReference: v1.LocalObjectReference{
+								Name: psSecret.Name,
+							},
+						},
+					}})
+			}
```
filariow, line 465: ⚠️ this should be superseded by next comment

this can be simplified
```suggestion
			e := v1.EnvVar{
				Name: e.Name,
				ValueFrom: &v1.EnvVarSource{
					SecretKeyRef: &v1.SecretKeySelector{
						Key: e.Key,
						LocalObjectReference: v1.LocalObjectReference{
							Name: psSecret.Name,
						},
					},
				},
			}
			if checkIfEnvPresentOrNot(c.Env, e) {
				c.Env = append(c.Env, e)
			}
```

However, if the EnvVar is already set but not up-to-date, it won't be updated.
I'd propose to introduce a new function named `createOrUpdateEnvVar(c *v1.Container{}, ev v1.EnvVar)` that adds the EnvVar if it does not exist, or update that if it exists.

controllers/agents/app/servicebinding_controller.go
```diff
@@ -501,11 +528,27 @@ func (r *ServiceBindingReconciler) removeVolumeMountFromContainer(ctx context.Co
...
+			var removeEnv bool
+			for _, bindingEnv := range sb.Spec.Env {
+				if env.Name == bindingEnv.Name {
+					if i < (len(c.Env) - 1) {
+						removeEnv = true
+					} else {
+						c.Env = c.Env[:i]
+					}
+				}
+			}
```
filariow, line 540: what if an environment variable is removed from the ServiceBinding?
I think that if we just remove all the EnvVars that are referencing the ServiceEndpointDefinition Secret before L443 would work better. WDYT?
The ones defined in the new generation of the ServiceBinding will be added again at L443.

this should also supersede the previous comment

filariow: may we also add this scenario in tests in this PR?

filariow (changes requested):
some feedback

controllers/agents/app/servicebinding_controller.go
Kartikey-star: We are only referring the secret in the field. And we do not allow sci to be updated in the service claim. So once the env is created , we won't need to update it's keys. Values would be updated in the secret itself. WDYT?

filariow: The user may change the secret key referenced by the EnvVar. In this case the change won't be persisted. I'd propose to use the approach described in the next comment for a more solid approach

controllers/agents/app/servicebinding_controller.go
Kartikey-star: If we remove the environment variable, everytime before the process starts, we might result in failing pods even when it is not required right.
And we might have an application which has environment variables other than what is projected by service binding.

filariow: Finally, the idea is just to remove the variables from the Service Binding's Secret, not all the variables

Kartikey-star: No, i mean the env variables not decalred by service binding specification

filariow: We should just remove variables that are pointing to the service Binding's secret. The other ones (user defined) should be left there

Kartikey-star: `I think that if we just remove all the EnvVars that are referencing the ServiceEndpointDefinition Secret before L443 would work better. WDYT?`... OK so you were suggesting to just remove all the env variables which match with the service binding spec, I interpreted that you menat to equate container env to nil

filariow (approved):
lgtm

PR #233, opened 2023-07-13
readme: add status badges

sadlerap:
We should add a few status badges for commonly-referenced materials, such as test results and community forums.  This is fairly standard practice in open-source communities (for example, check out [argocd's readme](https://github.com/argoproj/argo-cd/blob/master/README.md)).


filariow (approved):
LGTM

PR #238, opened 2023-07-17
implement registered service healthchecks

sadlerap:
This implements healthchecks for registered services.  A few important notes:
- The state of registered services now defaults in the CRD to the new state "Unknown", rather than "Available".  If a healthcheck is undefined or removed, then the service immediately becomes available, which is in line with old behavior.  However, if a healthcheck is defined, we don't want to transition it to "Available", since we haven't run the healthcheck yet.
- A new field, "minutes", has been added to the healthcheck's container field.  It represents how many minutes should be between each check. This is largely introduced for testing purposes, since the planned default of 5 minutes makes acceptance testing a pain.  Its value is restricted from 1 to 60.
- Healthchecks time out after thirty seconds.  We need to have some kind of timeout mechanism, since providing users a way to run processes (even if their unpriviledged) has a potential for abuse.  Hard-coding it to 30 seconds is fairly reasonable default, but we might want to revisit how we decide this in the future.


filariow (changes requested):

controllers/registeredservice_controller.go
```diff
@@ -199,37 +377,35 @@ func (r *RegisteredServiceReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		err := r.handleHealthcheck(ctx, &rs)
 		if err != nil {
-			log.Error(err, "RegisteredService Status Failed")
 			return ctrl.Result{}, err
 		}
-
-		err = r.reconcileCatalogs(ctx, rs)
-
+	} else {
+		err := r.cleanupHealthchecks(ctx, &rs)
```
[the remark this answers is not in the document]
sadlerap: We call `cleanupHealthchecks` whenever there isn't a healthcheck defined.  That should handle the case where a healthcheck is removed, right?

api/v1alpha1/types.go
```diff
@@ -57,7 +57,13 @@ type HealthCheckContainer struct {
 	// Container image with the client to run the test
 	Image string `json:"image"`
 	// Command to execute in the container to run the test
-	Command string `json:"command"`
+	Command []string `json:"command"`
+	// Minutes between runs of the healthcheck container
+	//+optional
+	//+kubebuilder:default:=5
+	//+kubebuilder:validation:Minimum:=1
+	//+kubebuilder:validation:Maximum:=60
```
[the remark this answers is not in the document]
sadlerap: Fixed in bc4b028.

controllers/registeredservice_controller.go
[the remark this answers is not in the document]
sadlerap: I've added a more explicit check for this in 55626e7.

controllers/registeredservice_controller.go
```diff
@@ -161,12 +167,184 @@ func (r *RegisteredServiceReconciler) addServiceToCatalog(ctx context.Context, s
...
+	}
+
+	l.Info("Job status", "completed", completed, "failed", failed)
+	if completed {
+		// only keep it in 'Claimed' if the healthcheck succeeded
+		if rs.Status.State == primazaiov1alpha1.RegisteredServiceStateClaimed {
+			rs.Status.State = primazaiov1alpha1.RegisteredServiceStateClaimed
+		} else {
+			rs.Status.State = primazaiov1alpha1.RegisteredServiceStateAvailable
+		}
```
dperaza4dustbit, line 295: I don't get the complete logic you are checking if the state is claimed set it to claimed? Would it not be better to just check if state != claimed set it to available

controllers/registeredservice_controller.go
```diff
@@ -161,12 +167,184 @@ func (r *RegisteredServiceReconciler) addServiceToCatalog(ctx context.Context, s
...
+	} else if failed {
+		rs.Status.State = primazaiov1alpha1.RegisteredServiceStateUnreachable
+	} else {
+		rs.Status.State = primazaiov1alpha1.RegisteredServiceStateUnknown
+	}
+
+	return r.registerHealthcheck(ctx, rs)
+}
+
+func (r *RegisteredServiceReconciler) cleanupHealthchecks(ctx context.Context, rs *primazaiov1alpha1.RegisteredService) error {
```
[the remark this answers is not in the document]
dperaza4dustbit: This will be a good feature to support yeah, they could be experimenting with different images to use as healthchecks

controllers/registeredservice_controller_test.go
```diff
@@ -0,0 +1,237 @@
...
+			Entry(genTestName, v1alpha1.RegisteredServiceStateAvailable, v1alpha1.RegisteredServiceStateUnreachable, batchv1.JobFailed),
+			Entry(genTestName, v1alpha1.RegisteredServiceStateClaimed, v1alpha1.RegisteredServiceStateClaimed, batchv1.JobComplete),
+			Entry(genTestName, v1alpha1.RegisteredServiceStateClaimed, v1alpha1.RegisteredServiceStateUnknown, batchv1.JobSuspended),
+			Entry(genTestName, v1alpha1.RegisteredServiceStateClaimed, v1alpha1.RegisteredServiceStateUnreachable, batchv1.JobFailed),
+			Entry(genTestName, v1alpha1.RegisteredServiceStateUnreachable, v1alpha1.RegisteredServiceStateAvailable, batchv1.JobComplete),
+			Entry(genTestName, v1alpha1.RegisteredServiceStateUnreachable, v1alpha1.RegisteredServiceStateUnknown, batchv1.JobSuspended),
+			Entry(genTestName, v1alpha1.RegisteredServiceStateUnreachable, v1alpha1.RegisteredServiceStateUnreachable, batchv1.JobFailed),
+		)
+	})
+})
```
dperaza4dustbit, line 237: Nice

controllers/registeredservice_controller.go
```diff
@@ -161,12 +167,184 @@ func (r *RegisteredServiceReconciler) addServiceToCatalog(ctx context.Context, s
...
+							Containers:    []corev1.Container{container},
+							RestartPolicy: corev1.RestartPolicyNever,
+							SecurityContext: &corev1.PodSecurityContext{
+								RunAsUser:    &uid,
+								RunAsGroup:   &uid,
+								RunAsNonRoot: &t,
+								SeccompProfile: &corev1.SeccompProfile{
+									Type: corev1.SeccompProfileTypeRuntimeDefault,
+								},
+							},
```
dperaza4dustbit, line 222: Hum... How are we passing the secrets needed for the client to authenticate to the service? We are not only checking tcp connectivity user might want to check the actual service is up and running and credentials are accurate to use hence "available" state.

dperaza4dustbit (changes requested):
Great approach, good implementation so far

controllers/registeredservice_controller.go
sadlerap: Fixed in d5f2d55.

controllers/registeredservice_controller.go
sadlerap: I think I'll punt on implementing this for a separate PR.  This is definitely going to take a bit more work, and I'd rather not hold up this PR by implementing it.

controllers/registeredservice_controller.go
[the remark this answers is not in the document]
filariow: Oh right, I did not understood the code correctly. That looks right to me then

controllers/registeredservice_controller.go
dperaza4dustbit: you got it

dperaza4dustbit (approved):
/lgtm

controllers/registeredservice_controller.go
```diff
@@ -161,12 +167,182 @@ func (r *RegisteredServiceReconciler) addServiceToCatalog(ctx context.Context, s
...
+func (r *RegisteredServiceReconciler) handleHealthcheck(ctx context.Context, rs *primazaiov1alpha1.RegisteredService) error {
+	l := log.FromContext(ctx).WithValues("namespace", rs.Namespace, "registered service", rs.Name)
+	cronjob := batchv1.CronJob{}
+	err := r.Get(ctx, types.NamespacedName{Name: rs.Name, Namespace: rs.Namespace}, &cronjob)
+	if k8errors.IsNotFound(err) {
+		l.Info("creating registered service healthcheck")
+		err := r.registerHealthcheck(ctx, rs)
+		rs.Status.State = primazaiov1alpha1.RegisteredServiceStateUnknown
+		return err
+	}
```
filariow, line 254: what if it is another error? if it's a network error it may still be that the CronJob is not existing, also `cronjob` will be `nil` in the following

filariow (changes requested):

filariow (approved):
/lgtm

PR #239, opened 2023-07-20
Adding Workflow for agenda generator

dperaza4dustbit:
This will generate the next community call agenda right after the current community call is finished

### DIFF
--- a/.github/workflows/agenda.yaml
+++ b/.github/workflows/agenda.yaml
@@ -1,0 +1,64 @@
+name: Agenda Generator
+
+on: # yamllint disable-line rule:truthy
+  schedule:
+    - cron: '15 12 * * 1' # run every Monday at 12:15 PM UTC
+  workflow_dispatch:
+    inputs:
+      meeting_date:
+        description: 'Meeting Date'
+        required: false
+        default: ""
+
+    branches:
+      - main
+      - agenda_workflow
+
+permissions:
+  discussions: write
+
+jobs:
+  create-discussion:
+    name: discussion-gen
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Print discussion date
+      id: meeting-date
+      run: |
+        if [ -z "${{ github.event.inputs.meeting_date }}" ]
+        then
+          meet_date=$(python3 -c 'import datetime; print((datetime.datetime.now() + datetime.timedelta(days=7)).strftime("%m/%d/%Y"))')
+        else
+          meet_date=${{ github.event.inputs.meeting_date }}
+        fi
+        echo "meetdate=$meet_date" >> $GITHUB_OUTPUT
+    - name: Create a new GitHub Discussion
+      id: create-discussion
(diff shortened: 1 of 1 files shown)


dperaza4dustbit:
I realized that this github action works but it is not maintained. I see warnings  and checking issues, there are some trivial issues about deprecated NoteJS and Github set-outputs. Switching to calling Github API directly.

.github/workflows/agenda.yaml
```diff
@@ -0,0 +1,53 @@
+name: Agenda Generator
+
+on: # yamllint disable-line rule:truthy
+  schedule:
+    - cron: '15 12 * * 1' # run every Monday at 12:15 PM UTC
+  workflow_dispatch:
```
filariow, line 6: I'd suggest to add a manual input here for the date. We could use this to manually schedule a Primaza Community Call that's not on a Monday.

.github/workflows/agenda.yaml
```diff
@@ -0,0 +1,53 @@
...
+          mutation {
+            # input type: CreateDiscussionInput
+            createDiscussion(input: {
+                                      repositoryId: "${{ vars.REPO_ID }}",
+                                      categoryId: "${{ vars.CAT_ID }}",
+                                      body:
+                                        "Welcome to our community call. This call will be held on ${{ steps.meeting-date.outputs.meetdate }} at [8:30 AM EDT (UTC-4)](https://mytime.io/08:30/EDT).
+                                        Click on [Primaza Google Meet](https://meet.google.com/rha-ftwg-orr) to join us. You can also join the [Primaza Group](https://groups.google.com/g/primaza/members) to receive a recurring invite on your calendar.
+                                        Agenda Items are entered through comments. Feel free to add your agenda Items.",
+                                      title: "Community Call Agenda ${{ steps.meeting-date.outputs.meetdate }}"}) {
```
filariow, line 41: We could add some emojis to make this post more attractive. This may be helpful: https://github.com/ikatyang/emoji-cheat-sheet/blob/master/README.md

filariow (changes requested):

.github/workflows/agenda.yaml
dperaza4dustbit: good idea, for example for this first one it will be manual

dperaza4dustbit: What about now: https://github.com/dperaza4dustbit/primaza/discussions/30

.github/workflows/agenda.yaml
dperaza4dustbit: I did check it out https://github.com/dperaza4dustbit/primaza/discussions/30

.github/workflows/agenda.yaml
filariow: great

filariow (approved):
linter is complaining about yaml, other than that lgtm

filariow (approved):
lgtm

PR #244, opened 2023-07-25
Ignore out/ directory when linting yaml

sadlerap:
Anything yaml under out/ is likely generated or retrieved from somewhere; linting it isn't particularly useful.  I've run into issues with the cached cert-manager manifest locally.


filariow (approved):
lgtm

PR #245, opened 2023-07-25
docs: registered service healthchecks

sadlerap:
Add some docs for registered service healthchecks


docs/book/src/entities/registeredservice.md
```diff
@@ -45,49 +46,86 @@ Please note that exclusions have precedence over inclusions.
...
+Once a RegisteredService is claimed with a [ServiceClaim][ServiceClaim], its state will transition to `Claimed`.
+If the [ServiceClaim][ServiceClaim] is deleted, then the RegisteredService will transition from `Claimed` back to `Available`, making it available for re-use.
 
-There are some situations when the health-check will fail while the RegisteredService is in `Claimed` state.
-In those cases the RegisteredService will move to `Unreachable`.
-If, at a later time, the health-check passes then the controller will check if there is still a claim matching the RegisteredService and move the state back to `Claimed`.
-However, if there is not claim matching the RegisteredService the state will move to `Available`.
+If the health check fails while the RegisteredService is in `Claimed` state, the RegisteredService will transition to `Unreachable`.
+If, at a later time, the health check passes, then the controller will check if there is still a claim matching the RegisteredService and transition to `Claimed`.
+Otherwise, the RegisteredService the state will transition to `Available`.
```
filariow, line 108: @sadlerap  When a RegisteredService transitions from `Claimed` to `Unreachable` what happens to the ServiceBinding?
How can we differentiate a RegisteredService that comes from `Unreachable` and needs to go back to `Claimed` from one that needs to go back to `Available`?

filariow (commented):
great job, just some questions

docs/book/src/entities/registeredservice.md
sadlerap: My understanding is that we should have the service binding deproject and reproject bindings into the application, so it should try to pick up a different service from the catalog.  I don't know if that's implemented or not, though AIUI a service with a failing healthcheck does get removed from the catalog.

In other words, there's no way for a service to transition from `Unreachable` to `Claimed`.  It can only transition from `Unreachable` to itself (check fails), `Available` (check succeeds), or `Unknown` (check is indeterminate).

docs/book/src/entities/registeredservice.md
```diff
@@ -45,49 +46,86 @@ Please note that exclusions have precedence over inclusions.
 For example, if the list contains `!prod` but also includes `dev`, then `dev` is considered to be in the `!prod` set of environments and therefore redundant.
 If there is a third environment stage, then `!prod` would include both `stage` and `dev` even if they are not defined in the list explicitly.
 
+### Health checks {health checks}
```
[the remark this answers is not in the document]
sadlerap: Nope, I've worked around needing this now.

docs/book/src/entities/registeredservice.md
dperaza4dustbit: Healthchecks are tricky to take action on, we need to design this with minimum disruption to the customer. The healthcheck can fail for a few seconds due to connectivity issues that are transitional, they could fail for a few minutes during a short outage or they can fail outright cause the backend service is gone. I think we need to design this with the threshold value in minutes for who long we are willing to wait until the claim controller decides to find another service to match. What I think we can do here no is go from unreachable to available when there is claim for this registered service and go back to claimed if there is a claim for this registered service. This should be possible

PR #250, opened 2023-08-01
Add discovered Service info in RegisteredService's metadata

filariow:
When a RegisteredService is created by the ServiceAgent, the RegisteredService's metadata are enriched with info from the Service.


Kartikey-star (approved):
Looks good to me

PR #252, opened 2023-08-02
Show services claimed-by-labels

baijum:
Ref. https://issues.redhat.com/browse/APPSVC-1373


pkg/authz/permission_list_app.go
```diff
@@ -73,4 +73,68 @@ var AppPermissionList = []Permission{
 		Name:          "primaza:app:leader-election",
 		Verbs:         []string{"create", "patch"},
 	},
+	{
```
dperaza4dustbit, line 76: Wonder why all these are changed in this PR not related to permissions

baijum: There was some bug with code generation. I tried to generate it manually.
I sent a separate PR #285 to fix the issue. Please take a look.

sadlerap (approved):

PR #253, opened 2023-08-02
Use SecretVolumeSource in ServiceBinding

filariow:

### DIFF
--- a/controllers/agents/app/servicebinding_controller.go
+++ b/controllers/agents/app/servicebinding_controller.go
@@ -192,18 +192,18 @@ func (r *ServiceBindingReconciler) GetSecret(ctx context.Context, serviceBinding
 func (r *ServiceBindingReconciler) PrepareBinding(ctx context.Context, serviceBinding *v1alpha1.ServiceBinding, psSecret *v1.Secret, applications ...unstructured.Unstructured) error {
 	l := log.FromContext(ctx)
 
+	f := false
+	p := int32(0444)
 	volumeName := serviceBinding.Name
 	mountPathDir := serviceBinding.Name
-	sp := &v1.SecretProjection{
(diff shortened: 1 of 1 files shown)


Kartikey-star (approved):

Kartikey-star:
Looks good to me. But out of curiosity what is the advantage of using secretvolumesource instead of projectedvolumesource

filariow:
> Looks good to me. But out of curiosity what is the advantage of using secretvolumesource instead of projectedvolumesource

From docs it seems to me they are identical (`DefaultMode` can be expressed in `ProjectedVolumeSource`)

```go
// Adapts a secret into a projected volume.
//
// The contents of the target Secret's Data field will be presented in a
// projected volume as files using the keys in the Data field as the file names.
// Note that this is identical to a secret volume source without the default
// mode.
type SecretProjection struct {
	LocalObjectReference `json:",inline" protobuf:"bytes,1,opt,name=localObjectReference"`
	// items if unspecified, each key-value pair in the Data field of the referenced
	// Secret will be projected into the volume as a file whose name is the
	// key and content is the value. If specified, the listed keys will be
	// projected into the specified paths, and unlisted keys will not be
	// present. If a key is specified which is not present in the Secret,
	// the volume setup will error unless it is marked optional. Paths must be
	// relative and may not contain the '..' path or start with '..'.
	// +optional
	Items []KeyToPath `json:"items,omitempty" protobuf:"bytes,2,rep,name=items"`
	// optional field specify whether the Secret or its key must be defined
	// +optional
	Optional *bool `json:"optional,omitempty" protobuf:"varint,4,opt,name=optional"`
}
```
The advantages I see in using `SecretVolumeSource` are conciseness and readability

PR #279, opened 2023-08-10
Fix release process

mmulholla:
Fixed a couple of bugs:
- application and service agent images not add to control_plane manifest.
- "null" in creationTimestamp field service agent manifest cause by bug in kustomize

Also added a parameter to the workflow to enable prevention of  image creation which is useful for testing the workflow. 

Tested the new manifests created with primazactl and all was good.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,6 +10,10 @@ on:  # yamllint disable-line rule:truthy
         description: "Primaza release version"
         required: true
         default: "latest"
+      build-image:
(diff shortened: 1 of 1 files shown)


filariow (commented):
other then linter complaints, lgtm

sadlerap (commented):
Looks good, so long as the yaml linter fixes are applied.

filariow (approved):
lgtm

mmulholla:
@sadlerap @filariow fixed the linter issues - sorry was not aware of that check.

PR #281, opened 2023-08-10
Reduce log verbosity in Application Agent

filariow:


controllers/agents/app/servicebinding_controller_informers.go
```diff
@@ -167,7 +166,6 @@ func (r *ServiceBindingReconciler) prepareUpdateFunc(ctx context.Context, synced
 
 		applicationResource := future.(*unstructured.Unstructured)
 
-		l.Info(fmt.Sprintf("application resource %v", applicationResource))
```
sadlerap, line 170: Rather than printing the entire application resource, could we get the name and namespace out of the resource and log that?  I like the logging we do here, but the entire resource is too much IMO.

filariow: I agree, let me change that

sadlerap (approved):
lgtm

PR #307, opened 2023-08-17
ci: use go v1.21 in nightly checks

sadlerap:
This was causing nightly testing failures, see
https://github.com/primaza/primaza/actions/runs/5885546588/job/15962174393 for an example.


filariow (approved):
LGTM

